PR #186, opened 2014-01-10
WIP: shovel scripts

jaberg:


gw0:
Looks interesting, but wouldn't separating them into usual scripts (like `hyperopt-mongo-worker`, not *shovel*-based) make them easier to use?

Otherwise it would be better if the same format of database connection strings would be used as elsewhere (like for `MongoTrials` constructor, `hyperopt-mongo-worker` parameter), not separating it into database host, port, and name? Also, your database strings are sometimes hard-coded in the WIP.

maxpumperla:
closing due to inactivity, feel free to reopen if you think this is still relevant

PR #223, opened 2015-04-01
Ipyworking

willgroves:
This code re-adds parallel execution support using recent versions of the IPython Parallel interface.


jaberg:
Thanks for the changes @willgroves, I hope that they haven't already gone stale in the time since you first proposed them!

PR #225, opened 2015-04-26
Support python3

lucasb-eyer:
It passes `nosetest hyperopt`, but I wasn't able to run `nosetest hyperopt` in py2 since I don't have a py2-scipy environment available.

One thing to note is that I had to fix the incompatibilities in `pyll` by hand because I couldn't figure out how to make `setup.py` also run `2to3` on additional packages.

### DIFF
--- a/hyperopt/pyll/base.py
+++ b/hyperopt/pyll/base.py
@@ -952,7 +952,10 @@ scope.define_pure(operator.getitem)
 scope.define_pure(operator.add)
 scope.define_pure(operator.sub)
 scope.define_pure(operator.mul)
-scope.define_pure(operator.div)
+try:
+    scope.define_pure(operator.div)
+except AttributeError:
+    scope.define_pure(operator.truediv)
 scope.define_pure(operator.floordiv)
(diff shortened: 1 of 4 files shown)


shjb:
Looks great @lucasb-eyer thanks for this!

jaberg:
Looks great @lucasb-eyer thanks for this!

jseabold:
What version of Python 3 was this on? I see a lot of errors because things like

```
import fmin as fmin_module
```

get fixed to 

```
from . import fmin as fmin_module
```

and this imports from `__init__.py` which has already imported the fmin function. Other things like this don't work, on 3.4 at least, because `i` isn't in the scope.

```
[(i,) for i in range(5)]
del i
```

jseabold:
Shadowing the module name with the function name is tricky. It's not immediately obvious to me how to fix this in a backwards compatible way.

lucasb-eyer:
Sorry, I don't remember the version, but it was likely 3.3 or 3.4, but I agree it's odd I didn't see the problems you found.

Issue #235, opened 2015-08-25
nosetests failing


dallascard:
Following the instructions in the documentation for installation (clone, setup, nosetests), nosetests fails for me. The issue seems to be the import of rv_generic, which I don't think is part of recent versions of scipy (0.14 onward). The inheritance from uniform_gen and norm_gen are also no longer correct.

None of these seem to be necessary, so I will remove them in a new branch and create a pull request.

~~~~~~~~~~~~~~
WARNING: IPythonTrials is not as complete, stable
         or well tested as Trials or MongoTrials.
..............................................................................SS..................../Users/dcard/Tools/ML/hyperopt/hyperopt/plotting.py:163: RuntimeWarning: divide by zero encountered in log
  y = np.log(vals[label])
....E.................................................................
======================================================================
ERROR: Failure: ImportError (cannot import name rv_generic)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Applications/anaconda/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Applications/anaconda/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Applications/anaconda/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/dcard/Tools/ML/hyperopt/hyperopt/tests/test_rdists.py", line 5, in <module>
    from hyperopt.rdists import (
  File "/Users/dcard/Tools/ML/hyperopt/hyperopt/rdists.py", line 9, in <module>
    from scipy.stats.distributions import rv_generic
ImportError: cannot import name rv_generic

----------------------------------------------------------------------
Ran 170 tests in 74.391s

FAILED (SKIP=2, errors=1)
~~~~~~~~~~~~~~~~

dallascard:
resolved by merge

PR #236, opened 2015-08-25
removed obsolete import and references to old versions of scipy from rdist.py

dallascard:
See issue [#235](a href=https://github.com/hyperopt/hyperopt/issues/235)


jaberg:
If everything still works, great!

I'd agree that "for completeness" isn't a good enough reason to have useless classes that cause import failures.

PR #242, opened 2016-01-03
pymongo 3 support

wskish:
pymongo.Connection is deprecated so switch to similar
pymongo.MongoClient.

Set write_concern parameters at time of MongoClient creation.

Update find() parameter names that have changed:
spec -> filter
fields -> projection


gu1234:
Thanks! works for me

PR #244, opened 2016-01-13
now doing relative import for fmin

JonathanRaiman:
On Python 3 `fmin` works, but fails when using a `Trials` object:

``` python
import pickle
import time
from hyperopt import fmin, tpe, hp, STATUS_OK, Trials

def objective(x):
    return {
        'loss': x ** 2,
        'status': STATUS_OK,
        # -- store other results like this
        'eval_time': time.time(),
        'other_stuff': {'type': None, 'value': [0, 1, 2]},
        # -- attachments are handled differently
        'attachments':
            {'time_module': pickle.dumps(time.time)}
        }
trials = Trials()
best = fmin(objective,
    space=hp.uniform('x', -10, 10),
    algo=tpe.suggest,
    max_evals=100,
    trials=trials)

print(best)
```

With the current change, importing `fmin` within the `base.py` now works as intended.

### DIFF
--- a/hyperopt/base.py
+++ b/hyperopt/base.py
@@ -621,8 +621,8 @@ class Trials(object):
         # -- Stop-gap implementation!
         #    fmin should have been a Trials method in the first place
         #    but for now it's still sitting in another file.
-        import fmin as fmin_module
-        return fmin_module.fmin(
+        from .fmin import fmin
+        return fmin(
             fn, space, algo, max_evals,
             trials=self,
             rstate=rstate,


jaberg:
is this fixed now? can someone with py3 set up comment?

bethke:
I have a py3 set up and this does work for me.

bethke:
I should clarify that the pull request on Jul 29 did not fix the problem but the solution above does fix the problem.  Currently I pip install from the master and then add this modification.

maexlich:
I also applied the changes to my py3 install and it seems to run without problems

maxpumperla:
We've addressed a little refactor in #272 that also includes this fix. Note that the above suggestion does not fix problems with distributed mode (due to how pymongo3 writes to mongo)

maxpumperla:
@maexlich If you don't mind I will close this PR.

Issue #245, opened 2016-01-15
Installation does not install dependencies


grisaitis:
Hi,

I'm finding that dependencies aren't being installed when I install hyperopt. I have tried installing with both pip and setup.py. 

Is this the intended behavior?

I see that setup.py mentions dependencies in `setuptools.setup` with `install_requires`. None of them are installed when I try installing with either `pip install hyperopt` or `git clone https://github.com/hyperopt/hyperopt.git` and then `(cd hyperopt && python setup.py install)`. 

I am using clean virtualenvs with Python 2.7.6. 

Thanks,
William

tounnas:
Hi,

With anaconda it's quite easy :

*conda install -c https://conda.anaconda.org/auto
<https://conda.anaconda.org/auto> hyperopt*

But if you whant to install it with pip,, it's quite tedious, you have to
install packages like pymongo, bson and others.. :S

Good luck,
Tom


2016-01-15 16:38 GMT+01:00 William C Grisaitis <notifications@github.com>:

CraigKelly:
@grisaitis, I just submitted a pull request to fix this. However, I have to admin that I generally don't use pip to install numpy and scipy. I install them at the OS level (I generally use Ubuntu), create my virtualenv with --system-site-packages, and then allow pip to install everything else. On a system with numpy and scipy already installed, this worked for me after I made the changes in my pull request:

```
virtual --system-site-packages env
. env/bin/activate
pip install ../hyperopt  # was in a separate test directory
python test.py
```

where test.py contained a simple example from one of the sample notebooks:

```python
from hyperopt import hp, fmin, rand
space = hp.choice('a', [-1, hp.uniform('b', 0, 1)])
best = fmin(fn=lambda x: x, space=space, algo=rand.suggest, max_evals=100)
print best
```

If you really need fully isolate your virtualenv, I've found that you'll generally do better with installing numpy and scipy separately instead of automatically. In this case your script could look something like:

```
virtualenv env
. env/bin/activate
pip install numpy
pip install scipy
pip install hyperopt
```

shahlaebrahimi:
I would appreciate if you could let me know how to install hyperopt on windows using anaconda?In fact, I am using anaconda on windows 10 so I tried this [instruction ](https://anaconda.org/jaikumarm/hyperopt) as well as the above instruction to install it as it shows below:

```
[...]

PR #246, opened 2016-01-15
Dependencies should be install by `pip install hyperopt`

CraigKelly:
I had the same problem as issue #245 

The issue was that install_requires needs a list, not an iterator as returned by `reversed`


jaberg:
Thanks!

PR #259, opened 2016-07-04
 available version for python2.7.11 and python3.5

sloth2012:
I found this version is available version in kaggle commuity.
the official version doesn't work now in py3.5 or py2.7.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -28,4 +28,3 @@ This project has received support from
 
 
 
-## this is bugs fix version in zrf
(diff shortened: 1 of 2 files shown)


jaberg:
Thanks!

sloth2012:
@ymtsang yes, you are right. I just notice that. I'm sorry about that. I will make a PR again. Thanks.

sloth2012:
@ymtsang I see that the fmin has been import in __init__.py. So this may be work.

Issue #253, opened 2016-03-26
hp.quniform giving float values for integer range.


muditbac:
When using hyperopt for optimizing XGBoost it throws error.
`XGBoostError: b"Invalid Parameter format for max_depth expect int but value='4.0'"`

I have set range as `{'max_depth': hp.quniform('max_depth', 1, 13, 1),}`

erlendd:
I had the same issue recently, but it's easily resolved by replacing hp.quniform by hp.choice('max_depth', np.arange(1, 13+1, dtyle=int).

yassineAlouini:
I had the same problem as @muditbac. As suggested by @erlendd, the `hp.choice('max_depth', np.arange(1, 14, dtype=int))` seems to solve the issue. Thanks for the help!

shahlaebrahimi:
I had the same problem so I used the suggested approach by @yassineAlouini and @erlendd. It solves the issue just in some iteration so again that error is reported.
I would greatly appreciate if you could let me know how to deal with it?
regards,

shahlaebrahimi:
Some part of the error is:
```
.637468	eval-mlogloss:0.67373
Multiple eval metrics have been passed: 'eval-mlogloss' will be used for early stopping.

Will train until eval-mlogloss hasn't improved in 10 rounds.
[1]	train-mlogloss:0.434509	eval-mlogloss:0.487707
[2]	train-mlogloss:0.395639	eval-mlogloss:0.462347
[3]	train-mlogloss:0.390629	eval-mlogloss:0.468641
[4]	train-mlogloss:0.355771	eval-mlogloss:0.420497
[5]	train-mlogloss:0.355757	eval-mlogloss:0.419786
[6]	train-mlogloss:0.355836	eval-mlogloss:0.418852
[7]	train-mlogloss:0.351659	eval-mlogloss:0.415997
[8]	train-mlogloss:0.351758	eval-mlogloss:0.415255
[9]	train-mlogloss:0.351417	eval-mlogloss:0.417624
[10]	train-mlogloss:0.351384	eval-mlogloss:0.418595
[11]	train-mlogloss:0.346445	eval-mlogloss:0.418504
[12]	train-mlogloss:0.346458	eval-mlogloss:0.425383
[13]	train-mlogloss:0.342745	eval-mlogloss:0.421794
[14]	train-mlogloss:0.343302	eval-mlogloss:0.417118
[15]	train-mlogloss:0.34298	eval-mlogloss:0.426841
[...]

PR #260, opened 2016-07-17
support uniformint

fukatani:
This PR is suggestion for supporting `uniformint` and `quniformint`
for integer hyper parameter such as `max_depth` of XGB.
(See #253 )

I know same function is achievable by using `hp.choice`, this PR can be declined.

### DIFF
--- a/hyperopt/hp.py
+++ b/hyperopt/hp.py
@@ -8,7 +8,9 @@ from pyll_utils import hp_randint as randint
 from pyll_utils import hp_pchoice as pchoice
 
 from pyll_utils import hp_uniform as uniform
+from pyll_utils import hp_uniformint as uniformint
 from pyll_utils import hp_quniform as quniform
+from pyll_utils import hp_quniformint as quniformint
 from pyll_utils import hp_loguniform as loguniform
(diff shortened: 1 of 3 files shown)


maxpumperla:
@jaberg are we interested in this? I think it's very valuable! @fukatani would you mind updating the branch so we can consider merging? Thanks!

jaberg:
I don't understand this PR and I wonder if @fukatani is misunderstanding quniform. quniform can already be used to truncate uniform draws to e.g. integer values, and it returns them as floating point numbers. It would be reasonable to further call scope.int on the return value from such a quniform call if the quantization value was an integer - is that the intent here? It isn't exactly what's written.

fukatani:
I'm sorry I misunderstand `quniform` and I want to merge only `uniformint`.
`quniformint` has no sense.

I think `uniformint` is useful in some cases where `integer` hyper parameters are only allowed.

I merged and resolved conflict.
I appreciate if you merge it.

jaberg:
How about implementing uniformint as a scope.int cast of hp_quniform with q hard-coded to be 1.0 ?

fukatani:
Thanks! I did it.

PR #262, opened 2016-07-29
Fix module not callable error

lishen:
The current version contains a bug that will trigger a "module not callable" error when line 625 in base.py is called. Unfortunately, this bug seems to be incorporated into several releases of hyperopt on anaconda.org too. This has prevented packages such as hyperopt-sklearn from running properly.

This PR is also related to PR #244, from which we can see that the older version of the code was correct. So the bug was only recently introduced.


jaberg:
thanks for taking the time to submit a patch.

PR #269, opened 2016-09-22
Unified usage of dill if available and fixed issues with lib2to3

### LINKED ISSUE #268: Usage of lib2to3 causes problems with pickle/dill
As the current version of hyperopt uses lib2to3 during installation, some files are changed at installation time.
This for example makes it impossible to use dill on a python 3 installation, as the import statement is not rewritten correctly:

Original fmin.py:

```
try:
    import dill as cPickle
except ImportError:
    import cPickle
```

after lib2to3:

```
try:
    import dill as cPickle
except ImportError:
    import pickle
```

This could of course be circumvented by the not very elegant statement

```
try:
    import dill as pickle
except ImportError:
    import cPickle as pickle
```

which would result in the strange but valid statement `import pickle as pickle` on python3 installations without dill.

maexlich:
Now dill is always used if available. Also this fixes #268 .

Can somebody test this on a py2.7 install?
From my perspective I see no potential problems, but a test run would be nice just to be sure.


maxpumperla:
We're using `six` to import py2/3 compatible pickling on master now. The logging for Mongo is still very interesting, though. Would you mind updating the branch @maexlich ?

maexlich:
@maxpumperla done

michaelmior:
The title of this PR is confusing since the commit seems to have nothing to do with dill...

PR #275, opened 2016-11-21
add future package to setup.py

### LINKED ISSUE #273: ImportError: No module named builtins (Python 2.7)
Running `pip install future` fixed this error for me.
For compatibility with Python2.7, I think the package `future` should be added to the `install_requires` in `setup.py` (hopefully this is also compatible with Python3?).

Note that `nosetests` also fails without matplotlib, but I'm not sure adding matplotlib as a dependency makes much sense.

maxpumperla:
Little patch to add `future` for python 2.x, closes #273

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,6 +144,6 @@ setuptools.setup(
     keywords='Bayesian optimization hyperparameter model selection',
     package_data=package_data,
     include_package_data=True,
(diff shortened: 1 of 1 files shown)


maxpumperla:
@jaberg ran this on virtualenvs with `nosetests` and `pytest`, both seem fine.

PR #290, opened 2017-02-13
Document the rather tricky code in tpe.py

johnflux:
This code is really difficult to understand.

I'd like to clean it up, but I thought it would be safer to at least document what is going on.


jaberg:
Thanks for writing this up. I don't know if even I can remember how TPE works, having these docs should help.

PR #309, opened 2017-06-05
fix mongo password authentication 

csanquer:
Hi,

With this PR, I fixed the mongo DB user/password authentication and authentication database parameter.

### DIFF
--- a/hyperopt/mongoexp.py
+++ b/hyperopt/mongoexp.py
@@ -171,16 +171,7 @@ class ReserveTimeout(Exception):
 
 
 def read_pw():
-    username = 'hyperopt'
-    password = open(os.path.join(os.getenv('HOME'), ".hyperopt")).read()[:-1]
-    return dict(
-        username=username,
-        password=password)
-
-
-def authenticate_for_db(db):
-    d = read_pw()
-    db.authenticate(d['username'], d['password'])
+    return open(os.path.join(os.getenv('HOME'), ".hyperopt")).read()[:-1]
 
 
 def parse_url(url, pwfile=None):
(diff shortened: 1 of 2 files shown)


abiteboul:
@jaberg Could we include this simple fix in milestone https://github.com/hyperopt/hyperopt/milestone/3 ?

jaberg:
I don't recall this is tested, how should it be tested? Does anyone have an idea how to set up Travis to do this?  cc @maxpumperla @michaelmior

jaberg:
I imagine this breaks compatibility with some old versions of Mongo, could you provide info on when Mongo started working this new way?

michaelmior:
AFAICT, it's pymongo which has changed, not MongoDB. If that's the case, I don't think we need to worry about backwards compatibility. People who want to use a new release of hyperopt will just need to install a new pymongo.

jaberg:
Someone reported recently that they tried running all tests, including
mongo tests, and they all passed.

Is it the case that no test performs authentication? Or maybe those tests
were run with an old version of pymongo?

How might we test authentication, see that this change works, and catch
similar future API regressions?

michaelmior:
I haven't looked at the tests currently configured, but if we have a test that uses authentication on Travis, this shouldn't be a problem.

abiteboul:
I don't think the current tests use any authentication. Actually, my understanding is that they do not use the MongoDB service as it is configured in `.travis.yml`, but rather generate a temporary database locally: https://github.com/hyperopt/hyperopt/blob/master/hyperopt/tests/test_mongoexp.py#L56

As it is now, the Travis build succeeds after:

abiteboul:
I don't think the current tests use any authentication. Actually, I think they do not use the MongoDB service as it is configured in `.travis.yml`, but (if I understand correctly) rather generate a temporary database locally: https://github.com/hyperopt/hyperopt/blob/master/hyperopt/tests/test_mongoexp.py#L56

As it is now, the Travis build succeeds after rebasing this merge request on the current `master`, even if we completely remove the MongoDB service in `.travis.yml`.

Personally, I feel like this is OK for the current release, as it adds authentication support while not breaking anything. Ideally, writing specific tests for MongoDB authentication would require a significant refactoring of `test_mongo.exp` in order to use the Travis CI service (probably using environment variables) rather than starting a new DB.

jaberg:
Thanks for thinking about this @abiteboul I agree with you.

If Travis passes this, let's merge it.

michaelmior:
This should probably be rebased on top of master. Tests are failing because the Travis config isn't in this branch.

csanquer:
The branch has been rebased on top of master.

PR #311, opened 2017-06-19
Added timing constraint

bacalfa:
Only having `max_evals` as a stopping criterion may lead to very long solving times.

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -56,6 +56,7 @@ class FMinIter(object):
                  max_queue_len=1,
                  poll_interval_secs=1.0,
                  max_evals=sys.maxsize,
+                 max_time=None,
                  verbose=0,
                  ):
         self.algo = algo
@@ -68,7 +69,10 @@ class FMinIter(object):
         self.poll_interval_secs = poll_interval_secs
         self.max_queue_len = max_queue_len
         self.max_evals = max_evals
+        self.max_time = max_time
         self.rstate = rstate
+
+        self.start_time = time.time() # Start clock
 
         if self.async:
             if 'FMinIter_Domain' in trials.attachments:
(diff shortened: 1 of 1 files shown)


jaberg:
Hi @bacalfa I wonder if this functionality could be delivered with better documentation.
The fmin() function can be used in a blocking style, but as you have found it's more flexible not to use it in a blocking style. It's more empowering to think about each fmin call as adding some N function evaluations to the trials object. N is max_evals - len(trials).

jaberg:
In this style you can check for wall time thresholds in an outer loop, but you can also check how your experiment is going as well, so if e.g. performance is good enough and your patience is up, then just don't call fmin() any more.

bacalfa:
Hi, James. I'd certainly be open to a tighter control of the wall time. :) I found that the solution I implemented has worked well for my purposes. However, I didn't try it with the "trials" approach, so not sure how well it works there. What would you suggest?

jaberg:
1. The first thing is to get a trials object explicitly
2. Call fmin(fn, trials, max_evals=5)
3. check the time, if it's going okay, then call fmin(fn, trials,
max_evals=10) to step 5 evals further
4. as before, then call fmin(fn, trials, max_evals=15) to step 5 evals
further yet

Maybe if the evals get longer as you go, you try 3 at a time, or 1.

Hope that helps?

On Mon, Aug 21, 2017 at 8:26 PM, Bruno Abreu Calfa <notifications@github.com

bacalfa:
James, I sent another push request. Is this closer to what you had in mind? Please let me know if you have any specific requests for changes. Thank you.

bacalfa:
Something isn't quite right... not sure what it is exactly. Again, I appreciate comments. :)

marctorrellas:
@bacalfa @jaberg what's the state of this? there's a similar PR here https://github.com/hyperopt/hyperopt/pull/533

bacalfa:
@marctorrellas Those commits are 2 years old... I think they were working for my purposes, but maybe they can be improved. I see that there's another proposed solution in #533. It's up to you to adopt that.

marctorrellas:
> @marctorrellas Those commits are 2 years old... I think they were working for my purposes, but maybe they can be improved. I see that there's another proposed solution in #533. It's up to you to adopt that.

Don't have any preference for either solution, just trying to unblock this. I think it's a nice feature that would be nice to have in next version. Given that you have conflicts, this PR cannot be merged (if it aligns with what @jaberg wanted, which I'm not sure it does or not). Therefore, I would say a) try to fix it, so it can be reviewed again, or b) close the PR

PR #319, opened 2017-08-21
Update toposort to work with networkx 2.0

### LINKED ISSUE #318: Incompatible with networkx 2.0
In networkx 2.0 they've replaced all their list-based API calls with generators. See https://networkx.github.io/documentation/development/reference/release_2.0.html

I suggest a dependency on networkx<2.0, or actually, there's only one use of networkX, I'll just submit a pull request.

davidsmf:
Fixes #318 - networkx have changed their interface to return a generator rather than a list, so the assertion was failing when it indexed the list. Create a list before returning - this keeps the toposort interface consistent. Also documented that toposort returns a list.

### DIFF
--- a/hyperopt/pyll/base.py
+++ b/hyperopt/pyll/base.py
@@ -703,7 +703,7 @@ def dfs(aa, seq=None, seqset=None):
 
 def toposort(expr):
     """
-    Return apply nodes of `expr` sub-tree in topological order.
+    Return apply nodes of `expr` sub-tree as a list in topological order.
 
     Raises networkx.NetworkXUnfeasible if subtree contains cycle.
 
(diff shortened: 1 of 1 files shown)


jaberg:
Awesome, thanks for the quick refresh for networkx 2.0.

vertti:
Why are we not getting a new release?

juharris:
`pip install --upgrade git+git://github.com/hyperopt/hyperopt.git`

ivan-marroquin:
It seems that the fixed made available on feb 12, 2017 was not added to the official version (installed using pip install hyperopt).

I just installed anaconda 3.4.1 on a windows machine - which includes networkx 2.1. 

Then, I got the same error as mentioned above in the fmin function.

My solution consisted of:
a) run "conda uninstall networkx"
b) run "conda install networkx=1.11"

PR #366, opened 2018-02-08
Fix mongo auth + pin down random tests

jaberg:
Trigger tests for #309

### DIFF
--- a/hyperopt/mongoexp.py
+++ b/hyperopt/mongoexp.py
@@ -171,16 +171,7 @@ class ReserveTimeout(Exception):
 
 
 def read_pw():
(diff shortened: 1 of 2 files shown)


jaberg:
Travis found another random test that sometimes fails - apparently kstest draws samples if you pass a callable first arg 
https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.kstest.html

Suggested fix is to go through test_rdists and replace kstest(callable, ...) with kstest(sample, ...) so we can control the random seed.

jaberg:
Replacing with #367

PR #372, opened 2018-02-21
do not log the MongoDB password in plain text

tgpfeiffer:
I think from a security point of view it makes sense to not log the password in plain text.

### DIFF
--- a/hyperopt/mongoexp.py
+++ b/hyperopt/mongoexp.py
@@ -222,7 +222,8 @@ def parse_url(url, pwfile=None):
             password = None
     else:
         password = tmp.password
-    logger.info('PASS ***')
+    if password is not None:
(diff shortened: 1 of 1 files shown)


tgpfeiffer:
Can I do anything to get this merged soon? I think the PR is rather straightforward to review.

michaelmior:
This logging was explicitly added for debugging purposes. Logging completely since logging `****` doesn't provide any useful information. It would at least be good to log whether a password was provided.

tgpfeiffer:
I see. So what do you think about something like:
```python
if password is not None:
  logger.info('PASS ***')
```

michaelmior:
That should work :)

tgpfeiffer:
@michaelmior Fixed, could you please have another look? Thank you!

PR #390, opened 2018-04-16
Accept raw Mongo Urls

andresespinosapc:
I allow to pass a raw mongo url and pass it directly to `MongoClient`, because you don't support `mongo+srv` protocol, which I needed to use. For this purpose, I had to remove `as_mongo_str` on `mongoexp.py` at line 1251, but I don't see any disadvantages of doing that, besides you have to update the documentation.

### DIFF
--- a/hyperopt/mongoexp.py
+++ b/hyperopt/mongoexp.py
@@ -200,13 +200,13 @@ def parse_url(url, pwfile=None):
     logger.info('HOSTNAME %s' % tmp.hostname)
     logger.info('PORT %s' % tmp.port)
     logger.info('PATH %s' % tmp.path)
-    
+
     authdbname = None
     if 'authSource' in query_params and len(query_params['authSource']):
         authdbname = query_params['authSource'][-1]
-        
+
     logger.info('AUTH DB %s' % authdbname)
-    
+
     try:
         _, dbname, collection = tmp.path.split('/')
     except:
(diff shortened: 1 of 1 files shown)


andresespinosapc:
It's weird, because the tests on the branch master are not passing in my computer neither :/ I don't know what is failing, help please

hyperopt/mongoexp.py
```diff
@@ -200,13 +200,13 @@ def parse_url(url, pwfile=None):
     logger.info('HOSTNAME %s' % tmp.hostname)
     logger.info('PORT %s' % tmp.port)
     logger.info('PATH %s' % tmp.path)
-    
+
```
michaelmior: Please keep whitespace changes out of PRs since it just pollutes the diff.

hyperopt/mongoexp.py
```diff
@@ -223,7 +223,8 @@ def parse_url(url, pwfile=None):
     else:
         password = tmp.password
     logger.info('PASS %s' % password)
-    port = int(float(tmp.port))  # port has to be casted explicitly here.
+    # If no port given, set it to de defaul (27017)
+    port = int(float(tmp.port)) if tmp.port else 27017  # port has to be casted explicitly here.
```
michaelmior: This seems to be unrelated to your proposed change

michaelmior:
This removes the ability to pass in a connection string without a protocol. Instead, I would suggest changing `as_mongo_str` to check for any protocol, adding `mongo://` as the default if none exists. This will allow `mongo+srv` to be used without breaking existing functionality.

michaelmior:
Is there now a way to use the `mongo+srv` protocol?

Issue #401, opened 2018-07-03
Passing previously generated random startup data


shup94:
Is there a way to pass my previously generated data as startup data? I've got a lot of combinations of hyperparameters from previous tests paired with the losses and would like to utilise this data for TPE.

Issue #395, opened 2018-05-28
Reusing trials from other similar hyperparameter configurations


ernestum:
I would like to do hyperparameter optimization on my algorithm and store the results using the MongoDBTrials Class. Later, I might add more hyperparameters or more choices in the choice nodes to the hyperparameter space. 

Now when tuning those parameters it would make sense to reuse the old trials to give some kind of good initial guess. Will this be possible? I guess changing the ranges of distributions is fine but will this work if I add more choices or entirely new parameters?

erniejunior:
I would prefer the second option because the first one might be confined to the specific trials implementation?

maxpumperla:
@erniejunior the solution @ptempczyk-nethone talks about has been merged in #403 already

PR #402, opened 2018-07-10
Function for adding custom starting points to be evaluated for fmin()

ptempczyk-nethone:
My pull request implements requests from issues #401 #395 . I have written a function to prepare Trials() object with custom hyperparameter points to evaluate at the beginning of the optimisation with fmin().

This function is the most non-invasive way to implement this idea but I can modify fmin() itself if you prefer this way.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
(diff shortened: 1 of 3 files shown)


maxpumperla:
@ptempczyk-nethone thanks for this. You were to quick for me to answer in the other issue :). Actually, I'd prefer to see an extension of `fmin` (in a backward compatible manner). Otherwise I'm afraid you'll get no visibility for this feature. Seems appropriate to add it to `fmin`, too.

ptempczyk-nethone:
Ok

ptempczyk-nethone:
I have created new pull request #403

PR #403, opened 2018-07-10
Parameter for adding custom starting points to be evaluated by fmin()

ptempczyk-nethone:
My pull request implements requests from issues #401 #395 . I have written a function to evaluate custom hyperparameter points to evaluate with fmin() at the beginning of the optimization.


hyperopt/fmin.py
```diff
@@ -24,7 +24,47 @@
     import dill as pickler
 except Exception as e:
     logger.info('Failed to load dill, try installing dill via "pip install dill" for enhanced pickling support.')
-    import six.moves.cPickle as pickler
```
maxpumperla: we need this indent

hyperopt/fmin.py
```diff
@@ -281,6 +322,11 @@ def fmin(fn, space, algo, max_evals, trials=None, rstate=None,
         for example if it is expected that `len(trials)` may be zero after
         fmin, and therefore `trials.argmin` would be undefined.
 
+    points_to_evaluate : list, default None
+        If None trials are evaluated normally. If dict with points to
+        evaluate is added the points are evaluated before optimisation starts.
```
maxpumperla: Can you maybe add an example and/or a more detailed description here what "points" look like? this is what users will ultimately read to use this. Like in your test below.

hyperopt/fmin.py
ptempczyk-nethone: fixed

hyperopt/fmin.py
ptempczyk-nethone: ok

ptempczyk:
@maxpumperla  Is there anything else I can correct in this PR?

maxpumperla:
@ptempczyk-nethone thanks for this, merging now.

PR #411, opened 2018-07-26
:bug: fix minor bug to mark new trial as running in fmin:serial_evaluate

jiayiliu:
Minor bug fix.  fmin serial_evalute() 
trial['state'] == base.JOB_STATE_RUNNING
should be assignment not comparison.

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -127,7 +127,7 @@ class FMinIter(object):
     def serial_evaluate(self, N=-1):
         for trial in self.trials._dynamic_trials:
             if trial['state'] == base.JOB_STATE_NEW:
-                trial['state'] == base.JOB_STATE_RUNNING
+                trial['state'] = base.JOB_STATE_RUNNING
                 now = coarse_utcnow()
                 trial['book_time'] = now
                 trial['refresh_time'] = now


stvreumi:
I think it is the same as  #287 . Why this bug is not merged?

Ingvar-Y:
@maxpumperla could you please merge this and close #287? This fix essentially does nothing because trial['state'] isn't used inside serial_evaluate() and gets overwritten inside
``` python
try:
    result = self.domain.evaluate(spec, ctrl)
except Exception as e:
    logger.info('job exception: %s' % str(e))
    trial['state'] = base.JOB_STATE_ERROR
    trial['misc']['error'] = (str(type(e)), str(e))
    trial['refresh_time'] = coarse_utcnow()
    if not self.catch_eval_exceptions:
        # -- JOB_STATE_ERROR means this trial
        #    will be removed from self.trials.trials
        #    by this refresh call.
        self.trials.refresh()
        raise
else:
    trial['state'] = base.JOB_STATE_DONE
    trial['result'] = result
    trial['refresh_time'] = coarse_utcnow()
```
Alternatively, we can just drop this line altogether.

maxpumperla:
@jiayiliu thanks for the fix and sorry for the insanely slow reaction time :)

Issue #410, opened 2018-07-26
MongoTrials and parallel search


ninnghazad:
Trying to get parallel search to work, i did like the wiki said, and have a few workers, a mongod and a job-creator. However there will only ever be one worker running at the same time. Starting multiple job-creators kind of gets the workers busy for one job, after they seem to get confused and it's back to one job at a time. 
what am i missing here? should i have a pool of fmins within one process? is there a parameter somewhere that i have to set to actually spawn enough jobs to satisfy all workers? or is this a bug and it should just work but doesn't?

ninnghazad:
tried some more setups, threads, processes, but it seems multiple fmin on the same exp_key wait on each others jobs. this might however just be the fault of keras/tensorflow and my own stuff. 
has anybody had a single process to serve a bunch of workers without waiting for all of them to finish before issuing new jobs?

richardliaw:
Maybe instead of the default HyperOpt parallelism mechanism, try using HyperOpt with [Tune](http://ray.readthedocs.io/en/latest/tune.html)? Tune uses HyperOpt to search but manages the asynchronous parallel execution for you and works with any deep learning framework.

```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

import ray
from ray.tune import run_experiments
from ray.tune.suggest import HyperOptSearch


def train_model(config, reporter):
    model = Model(config)  # create model with Hyperparameters
    for i in range(100):
        accuracy = model.train()
        reporter(mean_accuracy=accuracy)


if __name__ == '__main__':
    import argparse
    from hyperopt import hp

    ray.init()

    space = {
        'width': hp.uniform('width', 0, 20),
        'height': hp.uniform('height', -100, 100),
        'activation': hp.choice("activation", ["relu", "tanh"])
    }

    config = {
        "my_exp": {
            "run": "exp",
            "num_samples": 10,
            "stop": {
                "training_iteration": 100
            },
        }
    }
    # this wraps hyperopt TPE
    algo = HyperOptSearch(space, max_concurrent=32, reward_attr="neg_mean_loss")
    run_experiments(config, search_alg=algo)
```

ninnghazad:
@richardliaw thanks for the idea. i do however use multiple heterogeneous systems for hyperopting - which makes hyperopt's mongo-worker thing awesome for me. i think i will look into tune anyway - i could manually split the searchspace and distribute that - although that's kind of uncool.

[...]

Issue #192, opened 2014-02-13
Docs on how to run multiple trials simultaneously?


etrain:
I'm trying to use hyperopt to optimize several items in parallel with an external process. It seems that the MongoTrials object was designed to make this possible, but the code/docs inside mongoexp.py seems to be stale. (e.g. still reference hyperopt-mongo-search which doesn't seem to exist anymore)

Is there a good example anywhere of how to get fmin to dump out a few example points to mongo instead of just one at a time? What about writing results back to mongo when a run is done?

Thanks!

cuihenggang:
Hi, I have a similar question. I also hope to get a batch of suggestions to try, and then report their losses later. How do I do that?

I have figured out that I can get multiple suggestions with the tpe.suggest() function, but I couldn't figure out how to report back the loss. I think it probably has something to do with the "trials" argument passed to the tpe.suggest() function, but I don't know how to set it.

Do you have any insights on this?

Thank you!

ved93:
I am trying to use parallelization using below link
https://github.com/hyperopt/hyperopt/wiki/Parallelizing-Evaluations-During-Search-via-MongoDB

But when I run 1000-5000 iteration, the time taken by hyperopt increases as compared to a single thread. Also while parallelizing, core uses doesn't increase. I am using 8 core AWS system.
Can you please help -
1. I am not sure if mongo trials are using parallelization or not. If not how can I do it?
2. I need to run several hyperopt program in parallel. How can I do it? It seems with mongo trials its not possible because mongo worker uses all the cores.

ashay88:
I have created a pull request for this to re-enable _max_queue_len_ #417

PR #417, opened 2018-08-24
Update fmin.py to again support max_queue_len

ashay88:
Added a pass through for the max_queue_len parameter to FMinIter
Default value set to the earlier behaviour so should not affect the existing users.
Empirically I have found that increasing this value slightly helps to speed up parallel simulations which tend to lag sometimes on suggesting new trials.

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -251,7 +251,8 @@ def fmin(fn, space, algo, max_evals, trials=None, rstate=None,
          catch_eval_exceptions=False,
          verbose=0,
          return_argmin=True,
-         points_to_evaluate=None
+         points_to_evaluate=None,
+         max_queue_len=1
          ):
     """Minimize a function over a hyperparameter space.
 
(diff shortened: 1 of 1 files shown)


ashay88:
Related to this issue #192

maxpumperla:
@ashay88 can you please add a test for this feature? In particular, I want to guarantee we're backward compatible (is max_queue_len == 1 internally before this change?)

I think the travis failure is just time-out btw

abiteboul:
I think this should also fix https://github.com/hyperopt/hyperopt/issues/410

hyperopt/fmin.py
```diff
@@ -330,6 +331,11 @@ def fmin(fn, space, algo, max_evals, trials=None, rstate=None,
         Elements of this list must be in a form of a dictionary with variable
         names as keys and variable values as dict values. Example
         points_to_evaluate value is [{'x': 0.0, 'y': 0.0}, {'x': 1.0, 'y': 2.0}]
+        
+    max_queue_len : integer, default 1
+        Sets the queue length generated in the dictionary or trials. Increasing this 
+        value helps to slightly spped up parallel simulatulations which sometimes lag
```
nouiz: spped -> speed

ashay88:
@maxpumperla How and where should I add a test for the feature?
I am just passing through the value and setting it's default value to the default value(1.0) of the 2nd function. So currently nothing should break.
But later on if anyone changes the default value of the second function(FMinIter) things could show different results so a test could be useful.

PR #441, opened 2018-11-25
small fix for example of saving dot file

PhilipMay:
The example has a small bug:
`>>> open('foo.dot', 'wb').write(dot_hyperparameters(search_space()))`
The mode should not be 'wb' since it is a string that gets written.
This is right:
`>>> open('foo.dot', 'w').write(dot_hyperparameters(search_space()))`


maxpumperla:
thanks

PR #444, opened 2018-12-02
Preserve 'q' int type in stochastic expressions

brandonschabell:
The stochastic expressions do not currently function as described here: https://github.com/hyperopt/hyperopt/wiki/FMin#21-parameter-expressions. 

Specifically, the 'q' value types are lost. For example, hp.quniform(label='example', low=0, high=10, q=1) returns float values (e.g. 5.0) instead of integers. This leads to problems downstream when the results of the fmin() are required to be of an integer type.

This fix simple adds a cast to the rounding function, _round(uniform(low, high) / q)_, so the return type is naturally handled by the subsequent multiplication by q.

### DIFF
--- a/hyperopt/tests/test_rdists.py
+++ b/hyperopt/tests/test_rdists.py
@@ -23,6 +23,9 @@ try:
     )
 except ImportError:
     # XXX skip-tests instead or fix the import error, why would it fail?
+    
+    # -- This is failing because of a bug in the scipy.tests.test_continuous import.
+    # https://github.com/scipy/scipy/commit/c05c6b73a0cd16c4cc12a23e7ad72742dfeb42f1 will fix it with scipy 1.2.0
 
     def check_cdf_logcdf(*args):
         pass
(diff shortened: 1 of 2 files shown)


hyperopt/tests/test_rdists.py
```diff
@@ -152,6 +155,13 @@ def test_offgrid_float(self):
         assert qn.pmf(.99) == 0.0
         assert qn.pmf(-.99) == 0.0
 
+    def test_output_type_int(self):
+        result = quniform_gen(0, 10, 1).rvs()
```
yassineAlouini: One minor detail: is there a reason why you assign the thing you want to test to `result` in the `*_int` tests but 
test it directly in the `*_float` ones?

maxpumperla:
@brandonschabell beautiful! also, thanks for your comment on scipy 1.2.1, very helpful.

maxpumperla:
for smaller PRs like these you can just ping me. James is a little too busy right now, I'd only involve him for big design decisions (my judgement call).

brandonschabell:
@maxpumperla No problem, thanks!

Issue #454, opened 2019-01-09
Add a Progressbar with tqdm?


PhilipMay:
Hi,
what do you think about adding a progressbar to hyperopt?
It would show the progress to max_evals if it is given.
Maybe it is turned off by default and can be turned on by a parameter.
I would contribute it.

I think it could be added here:
https://github.com/hyperopt/hyperopt/blob/master/hyperopt/fmin.py#L238

I would use this implementation:
https://github.com/tqdm/tqdm

What do you think?

Ingvar-Y:
I'm not a contributor - just a user - but I would love tqdm progress bar. Hyperopt optimization takes a lot of time for my models so progress bar + finish time estimate would be a nice QoL addition.

PhilipMay:
So do you want to have an on/off switch for the progress bar?
And do you want to have it on or off by default?

PhilipMay:
> I'm not a contributor - just a user - but I would love tqdm progress bar. Hyperopt optimization takes a lot of time for my models so progress bar + finish time estimate would be a nice QoL addition.

I added the progressbar. See #454
It can not be turned off and is on by default. But I can add this if wanted.
The progressbar also shows the best (lowest loss).
Would be happy for feedback.

Philip

ninnghazad:
this would be great. but not sure about having it on by default - might mess up some stuff downstream.

maxpumperla:
@ninnghazad PR is merged, prog bar is off by default now.

PhilipMay:
@maxpumperla I do not see that the progbar is off by default now. Did you change anything or am I just blind?

See here: https://github.com/hyperopt/hyperopt/blob/master/hyperopt/fmin.py#L274

maxpumperla:
@PhilipMay sorry, lost track of our discussion. it's on by default, but can be turned off at any point. the way you implemented it, that should be fine.

PhilipMay:
Closing this since it is done IMO.

PR #455, opened 2019-01-10
progressbar with tqdm

PhilipMay:
I added the progressbar. See #454 
It can not be turned off and is on by default.
But I can add this if wanted.

The progressbar also shows the best (lowest loss).

Would be happy for feedback.

Philip

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ install:
   - pip install $IPYTHON
   - pip install --only-binary=numpy,scipy numpy scipy
   - python setup.py install
-  - pip install pytest pytest-cov pep8 pytest-pep8
+  - pip install pytest>=3.6 pytest-cov pep8 pytest-pep8
 script:
   - python -m pytest
 after_success:
--- a/hyperopt/base.py
+++ b/hyperopt/base.py
@@ -603,6 +603,7 @@ class Trials(object):
              pass_expr_memo_ctrl=None,
              catch_eval_exceptions=False,
              return_argmin=True,
+             show_progressbar=True,
              ):
         """Minimize a function over a hyperparameter space.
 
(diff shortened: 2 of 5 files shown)


PhilipMay:
I am working on it. Will tell you when I am done.
The distributed setup (mongo etc.) should work too. I will investigate it.

maxpumperla:
awesome, looking forward to it.

PhilipMay:
> btw, how does this look in distributed setup (mongo etc.)?

Well, for mongo you pass a `MongoTrials` object to `fmin` instead of a "normal" `Trials` object. See here: https://github.com/hyperopt/hyperopt/wiki/Parallelizing-Evaluations-During-Search-via-MongoDB#2-use-mongotrials

I am pretty sure that they are handled by the same loop that is inside my progress bar.

PhilipMay:
I am done with my work here. The python 2.7 build is still broken though. :-(
What do you suggest?

PhilipMay:
When should I extend the documentation? https://github.com/hyperopt/hyperopt.github.io
After release?

maxpumperla:
@PhilipMay I tested locally, works with 2.7. Let's fix travis after this PR. Same goes for docs.

Thanks for the contribution, please keep it up! :)

PhilipMay:
Was fun to implement it and discuss it here. ! :-)

trevorwelch:
Thankful for this! Great job!

uduse:
This is a nice feature to have, but unexpected output like this breaks doctests. Before this PR, I managed to silence hyperopt's output by setting the logger verbose level. However, it no longer works now. Since I also think having `show_progressbar=True` by default is a great idea, I suggest instead of using plain tqdm, [pass tqdm to the logger](https://github.com/tqdm/tqdm/issues/313) so that projects do not care about hyperopt's output get this nice feature, and projects fiddled with hyperopt's logger level have consistent output behavior.

PR #459, opened 2019-01-20
One-sided truncation in Gaussian Mixture Models for TPE 

jarednielsen:
I noticed a simple TODO while reading the source code, so thought I'd give a little contribution!


maxpumperla:
@jarednielsen thanks for this, and also taking care of handling the `float(None)` error.

Issue #458, opened 2019-01-18
Storing choice values by index instead of value?!


PhilipMay:
Hi,
here I have a short demo with using `hp.choice`:
https://gist.github.com/PhilipMay/a3bea840669acfba8293291dfe438f54

Objective is being called with `'y': 'y1'` and `'y': 'y2'`. But the return value of `fmin` is not storing the value for `y`but the index as you can see after the link.

IMHO this is inconsistent and the consequence is that I can not use the result of fmin to call the objective function.

Would be great if someone could have a look into this. Although I am aware of the fast that changing this might be a breaking change...

Philip

maxpumperla:
yeah, a lot of people stumble on this. we get recurring issues whose root cause is precisely this behaviour. I'd be very careful in changing this, though. As you said, it's a breaking change and we might do more harm than good.

Any ideas on how to pull this off by keeping backward compatibility? Maybe have a switch `return_indices` which is `True` by default and that you can turn off to return values? The problem with this is that it will be difficult to convey this new concept. How would we document this well enough and let people know?

maxpumperla:
Don't think so, but we can and should test it.

FlorisHoogenboom:
Wouldn't it be as simple to reuse the `return_agmin` argument in `fmin` ? This defaults to `True`, and if set otherwise nothing is returned by `fmin`. Wouldn't it be possible to simply return the value of `space_eval` called with the `argmin` found?

If you agree with this I would be more than happy to prepare a PR for this.

PhilipMay:
Here is a short summery of argmin:
```
    argmin : None or dictionary
        If `return_argmin` is False, this function returns nothing.
        Otherwise, it returns `trials.argmin`.  This argmin can be converted
        to a point in the configuration space by calling
        `hyperopt.space_eval(space, best_vals)`.
```

To me this somehow sounds very weird because `trials.argmin` should already be the point in configuration space. There should be no need to convert it again. But this might be the solution to my "problem". At least it is mentioned in the docstring.

FlorisHoogenboom:
I've make a PR since it is relatively simple to return the result of `space_eval` when `return_argmin` is set to `False`. This should be not that invasive since now it returns `None` in that case. Also, by default `return_agmin` is set to `True` so default behaviour won't change. This should resolve a lot of issues here.

maxpumperla:
see #463

Issue #431, opened 2018-10-05
hp.choice() unexpected results


r0f1:
Code:

```
from hyperopt import hp
from hyperopt import tpe
from hyperopt.fmin import fmin
from hyperopt import Trials

search_spaces = {
    'characters': hp.choice('characters', ["a", "b", "c", "d", "e", "f", "g"])
}

def objective(param):
    print(param)
    return 42

trials = Trials()
best = fmin(objective, search_spaces, tpe.suggest, max_evals=5, trials=trials)

print()
print(best)
```

Prints this result:

```
{'characters': 'e'}
{'characters': 'f'}
{'characters': 'f'}
{'characters': 'b'}
{'characters': 'g'}

{'characters': 4}
```

How come, that within the objective function I get a proper value for `characters` (that is, one character from the provided list of characters), but as output, called `best` here in this example, I get an index, rather than the character?

bjkomer:
I don't remember the exact reasoning for why it's done this way, but I think it's related to the fact that `hp.choice` in the general case can have nested search spaces as elements. The actual parameter being chosen is the index rather than the value, since the value can still vary even when the parameter is fixed.

Take a look at #384 for ways to access the value itself. There is a `hyperopt.space_eval` function to do this.

Although the question still remains, should this be the default return value instead? It seems like it would be more intuitive than getting an index.

maxpumperla:
see #463

PR #463, opened 2019-02-01
Change fmin to return the best parameter values if return_argmin is False

### LINKED ISSUE #412: Return found parameters in the same form as input
TL;DR `fmin` returns **flattened** dict of conditional parameter values that differs from the input format and the values passed to the `fn` function.

Example
-----------

Let me use a toy example, say that we want to find parameters of linear regression with "random" data:

```python
import numpy as np
import hyperopt
from hyperopt import fmin, tpe, rand, hp

x = np.linspace(-6, 6, 5000)
f = lambda x : 5 + x * 3 + np.random.normal(size=len(x), scale=2)

def opt_func(param):
    yhat = 0
    if param['b0']:
        yhat += param['b0']['b0_coef']
    if param['b1']:
        yhat += param['b1']['b1_coef'] * x
    if param['b2']:
        yhat += param['b2']['b2_coef'] * x**2
    return np.mean((f(x) - yhat)**2)
```

We define conditional parameters in the search space:

```python
search_space = {
    'b0' : hp.choice('b0', [False, { 'b0_coef' : hp.uniform('b0_coef', -50, 50) }]),
    'b1' : hp.choice('b1', [False, { 'b1_coef' : hp.uniform('b1_coef', -50, 50) }]),
    'b2' : hp.choice('b2', [False, { 'b2_coef' : hp.uniform('b2_coef', -50, 50) }]),
}
```

When we sample from it, we get the samples in the similar "conditional" format as input:

```python
hyperopt.pyll.stochastic.sample(search_space)
### {'b0': False, 'b1': {'b1_coef': -27.446828750200137}, 'b2': False}
```

But the output returned by optimizer differs:

```python
fmin(fn=opt_func, space=search_space, algo=rand.suggest, max_evals=500)
### {'b0': 0, 'b1': 1, 'b1_coef': 3.2529256424748496, 'b2': 0}
```

Why this is a problem?
-----------------------------

Well, this means that we need to define **different** function for unpacking the parameters when optimizing and another one for unpacking the "final" parameters.

Solution
----------

Use `hyperopt.space_eval` for the output as a default.

FlorisHoogenboom:
This PR adds support for returning the actual values that led to the optimal solution instead of the optimal arguments of the space. This is done by making `fmin` return the result of `hyperopt.space_eval` instead of `None` when the `return_argmin` argument to `fmin` is set to `False`.

This resolves #412, #431, #458.

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -362,13 +362,11 @@ def fmin(fn, space, algo, max_evals, trials=None, rstate=None,
     Returns
     -------
 
-    argmin : None or dictionary
-        If `return_argmin` is False, this function returns nothing.
-        Otherwise, it returns `trials.argmin`.  This argmin can be converted
-        to a point in the configuration space by calling
-        `hyperopt.space_eval(space, best_vals)`.
-
-
+    argmin : dictionary
+        If return_argmin is True returns `trials.argmin` which is a dictionary.  Otherwise
+        this function  returns the result of `hyperopt.space_eval(space, trails.argmin)` if there
+        were succesfull trails. This object shares the same structure as the space passed.
+        If there were no succesfull trails, it returns None.
     """
     if rstate is None:
         env_rseed = os.environ.get('HYPEROPT_FMIN_SEED', '')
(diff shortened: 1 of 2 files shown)


FlorisHoogenboom:
Hmm, build fails for 2.7, however that seems unrelated to the contents of this PR?

maxpumperla:
@FlorisHoogenboom thanks for this contribution. travis struggles with 2.7 for unrelated reasons, didn't find the time to get around this yet.

FlorisHoogenboom:
Okay, thnx. Good to know.

Could you maybe close #458 and #431? I see GitHub only closed the first issue listed automatically.

PR #501, opened 2019-05-14
Meaningful exception for all optimization trials failed

twolodzko:
When all optimization trials fail with status `base.STATUS_FAIL`, the code fails with "ValueError: attempt to get argmin of an empty sequence". This PR replaces the error with more meaningful exception, that is easier to catch and recognize.

**Example:**
```python
def objective(x):
    return {'status': STATUS_FAIL}

trials = Trials()
best = fmin(objective,
    space=hp.uniform('x', -10, 10),
    algo=tpe.suggest,
    max_evals=100,
    trials=trials)
```

### DIFF
--- a/hyperopt/base.py
+++ b/hyperopt/base.py
@@ -44,8 +44,8 @@ except ImportError:
 from . import pyll
 from .pyll.stochastic import recursive_set_rng_kwarg
 
-from .exceptions import (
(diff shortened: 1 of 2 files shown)


maxpumperla:
@twolodzko thanks for this addition!

PR #505, opened 2019-05-24
fix fileno error

jpizarrom:


maxpumperla:
@jpizarrom mind explaining this? the only mention of `fileno` I see is in your PR. Where do we need this?

afbarnard:
This is needed because running external programs currently fails when the progress bar is enabled because `DummyTqdmFile` doesn't have a `fileno` method which is called as part of `subprocess.Popen`. Do you want me to open a separate issue? Or should I just make a bug report here?

PR #506, opened 2019-05-25
fixing max_queue_len for mongotrials

jaikumarm:
fixes issue #410.

without the fix, the default max_queue_len=1 gets passed to fmin and its not able to queue more than 1 job at a time, with this fix I can generate multiple jobs in queue with mongotrials.


maxpumperla:
@jaikumarm thanks, good catch. could you please pin `networkx==2.2` in here:

https://github.com/hyperopt/hyperopt/blob/master/setup.py#L150

to fix build issues for Python2?

maxpumperla:
I know this isn't related to your change, but I don't want to merge a broken build. so apologies

jaikumarm:
Done. I saw the build failures, however did not want to presume how you wanted that fixed, in this PR or a different one. 

PS: thank you for maintaining this package, I rely on this package a lot, so appreciate your time and effort.

maxpumperla:
@jaikumarm thank you!

Issue #414, opened 2018-08-02
Parallel execution of hyperopt on Spark


asstergi:
Has anyone tried to adapt hyperopt to run in parallel on Spark?

I may need to do so and looking for a starting point.

maxpumperla:
I started doing this with elephas, but it may need some work

asstergi:
@maxpumperla I read on elephas that `right now, the only available optimization algorithm is random search`. Is this still true? Any reasons that TPE was more difficult to implement?

maxpumperla:
@asstergi that is still true. I just wanted to give you a hint on how this could be done. more importantly, elephas only supports keras models. you might want something else.

what you want to do is hijack spark for model and data distribution, compute trials on workers, collect them on master and then send out a new set of trials. The "problem" is that Spark can not easily be used as async job queue. The main paradigm in Spark is to set up an RDD or DataFrame and compute a single operation on distributed data in a synchronous fashion, i.e. your master will get the results when all workers are done. For TPE to work well you want to get back results asap so your new trial selection is based on recent info. To circumvent that, you'd need to set up some sort of parameter server that does this for you (elephas has an experimental HTTP server to do this). MongoDB is better suited for this kind of parallelization in a way, but I agree it would be nice to have a Spark integration for hyperopt.

How exactly do you envision this to go down? How do you want to set up your experiments?

Might be interesting as standalone project for the future. Can be done

asstergi:
@maxpumperla Thank you for all this information! I'm actually now working with Keras models but I definitely need something more generic.

I've quickly checked elephas code and as far as I understand, the code retrieves as many random hyper-parameter configurations as needed and distributes them to the workers to run each model (please correct me if I'm wrong).

I was thinking something along the lines of what you said. To set-up a parameter server and somehow make this work. I haven't thought a lot on this. I'm just checking the waters to see if there's anything that works out of the box.

maxpumperla:
that is correct. keep me posted on what you end up doing, very interesting topic

jkbradley:
@maxpumperla I work with Apache Spark and am interested in contributing this integration feature to hyperopt.  My team has done a bit of prototyping, and I think we can achieve reasonably high parallelism without a heavy-duty addition like a parameter server.

Do you have spare cycles to help review the work?  We can share more info either here or in a design doc.

Also @asstergi were you able to test this out?

Thanks both in advance!

CC @yifan-cao

asstergi:
[...]

PR #509, opened 2019-06-06
Add SparkTrials

hanyucui:
This PR integrates hyperopt with Apache Spark clusters.  The goal is to provide users a scalable distributed hyperopt implementation under the hood.  This is limited to hyperopt tasks which can run on a single machine, e.g., for scikit-learn.

A `SparkTrials` type is added to distribute trial runs across a Spark cluster. It uses concurrent Spark jobs for running asynchronous hyperopt tasks on Spark executors.

This PR is based on this [design doc](https://docs.google.com/document/d/1R-qMVOYxaLw3HdqMiTADQDZ5z7uInMR-x42HFCXu2-s/edit), which was discussed with @maxpumperla in #414.  Contributors include @WeichenXu123, @lu-wang-dl, @jkbradley, and the author.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,25 +7,25 @@ cache:
     - $HOME/.cache/spark-versions
 
 env:
-   global:
-      - HYPEROPT_FMIN_SEED=3
-      - SPARK_VERSION=2.4.3
-      - SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.6"
-      - SPARK_BUILD_URL="https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_BUILD}.tgz"
-      - SPARK_HOME=$HOME/.cache/spark-versions/${SPARK_BUILD}
-      - RUN_ONLY_LIGHT_TESTS=True
-      - NUMPY_VERSION=1.14.3
+  global:
+    - HYPEROPT_FMIN_SEED=3
+    - SPARK_VERSION=2.4.4
+    - SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
+    - SPARK_BUILD_URL="https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_BUILD}.tgz"
+    - SPARK_HOME=$HOME/.cache/spark-versions/${SPARK_BUILD}
+    - RUN_ONLY_LIGHT_TESTS=True
+    - NUMPY_VERSION=1.14.3
 matrix:
-   include:
-      - python: 2.7
-        env:
-          - IPYTHON="ipython[all]==5.4.1"
-      - python: 3.5
-        env:
-          - IPYTHON=ipython[all]
-      - python: 3.6
-        env:
-          - IPYTHON=ipython[all]
+  include:
+    - python: 2.7
+      env:
+        - IPYTHON="ipython[all]==5.4.1"
+    - python: 3.5
+      env:
+        - IPYTHON=ipython[all]
+    - python: 3.6
+      env:
+        - IPYTHON=ipython[all]
 
 services:
   - mongodb
@@ -35,7 +35,7 @@ before_script:
   - mongo mydb_test --eval 'db.createUser({user:"travis",pwd:"test",roles:["readWrite"]});'
 
 before_install:
- - ./download_spark_dependencies.sh
+  - ./download_spark_dependencies.sh
 
 install:
   - sudo apt-get remove ipython
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,31 @@
 dist: trusty
+
 language: python
 
+cache:
+  directories:
+    - $HOME/.cache/spark-versions
+
+env:
+   global:
+      - HYPEROPT_FMIN_SEED=3
+      - SPARK_VERSION=2.4.3
+      - SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.6"
+      - SPARK_BUILD_URL="https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_BUILD}.tgz"
+      - SPARK_HOME=$HOME/.cache/spark-versions/${SPARK_BUILD}
+      - RUN_ONLY_LIGHT_TESTS=True
+      - NUMPY_VERSION=1.14.3
 matrix:
    include:
       - python: 2.7
         env:
-          - IPYTHON="ipython[all]==5.4.1" HYPEROPT_FMIN_SEED=3
+          - IPYTHON="ipython[all]==5.4.1"
       - python: 3.5
         env:
-          - IPYTHON=ipython[all] HYPEROPT_FMIN_SEED=3
+          - IPYTHON=ipython[all]
+      - python: 3.6
+        env:
+          - IPYTHON=ipython[all]
 
 services:
   - mongodb
@@ -17,16 +34,20 @@ before_script:
   - sleep 15  # mongo takes time to start
   - mongo mydb_test --eval 'db.createUser({user:"travis",pwd:"test",roles:["readWrite"]});'
 
+before_install:
+ - ./download_spark_dependencies.sh
+
 install:
   - sudo apt-get remove ipython
   - pip install --upgrade pip
   - pip install --upgrade setuptools
   - pip install --upgrade wheel
   - pip install $IPYTHON
-  - pip install --only-binary=numpy,scipy numpy scipy
+  - pip install --only-binary=numpy,scipy numpy==$NUMPY_VERSION scipy
   - python setup.py install
   - pip install pytest>=3.6 pytest-cov pep8 pytest-pep8
+  - pip install wrapt
 script:
-  - nosetests
+  - ./run_tests.sh
 after_success:
   - coveralls
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -1,0 +1,415 @@
+from __future__ import print_function
+import numbers
+import threading
+import time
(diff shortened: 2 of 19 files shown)


jkbradley:
@maxpumperla Here's the holistic PR!  Some notes about it:
* **Pickling**: We switched from dill to cloudpickle, which handles closure capture better on clusters.  If this is an issue, we can figure out if we can make do with dill or use both.  (We tried dill initially but didn't get it to work on a real cluster.)
* **Trial cancellation**: We added this as a concept so that we could handle cases when Spark jobs were cancelled.  This was partly for the timeout parameter and partly in case a user killed some of the current Spark jobs.  This is the part which will require an OSS Spark change, though the code works with Spark 2.4.3 (but is less able to cancel all Spark jobs upon demand).
* ** best_trial**: We changed the semantics of this slightly but can revert the change if you're worried about user code breaking.  This was to avoid uses of best_trial within Hyperopt causing exceptions when some or all trials returned NaN.

Let us know what you think, and whether we should break this up for simpler review!

.travis.yml
```diff
@@ -1,14 +1,31 @@
...
 
+cache:
+  directories:
+    - $HOME/.cache/spark-versions
+
+env:
+   global:
+      - HYPEROPT_FMIN_SEED=3
+      - SPARK_VERSION=2.4.3
+      - SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.6"
```
maxpumperla: can we get a consistent indentation here?

hyperopt/tests/test_spark.py
```diff
@@ -0,0 +1,434 @@
+import contextlib
+import logging
+import unittest
+import tempfile
+import time
+import shutil
+
+from six import StringIO
+
+from pyspark.sql import SparkSession
```
maxpumperla: shouldn't we explicitly require `pyspark` as dependency in `setup.py`?

hyperopt/spark.py
```diff
@@ -0,0 +1,415 @@
...
+     - 'refresh_time': timestamp for last status update
+     - 'misc': includes:
+        - 'error': (error type, error message)
+     - 'book_time': timestamp for trial run start
+    """
+
+    asynchronous = True
+
+    # Hard cap on the number of concurrent hyperopt tasks (Spark jobs) to run. Set at 128.
+    MAX_CONCURRENT_JOBS_ALLOWED = 128
```
maxpumperla: just curious, why 128? Might be a good idea to add util functionality to change this, if needed.

hyperopt/spark.py
```diff
@@ -0,0 +1,415 @@
...
+        super(SparkTrials, self).__init__(exp_key=None, refresh=False)
+        self._spark_context = SparkSession.builder.getOrCreate().sparkContext if spark_session is None \
+                                  else spark_session.sparkContext
+        # The feature to support controlling jobGroupIds is in SPARK-22340
+        self._spark_supports_job_cancelling = hasattr(self._spark_context.parallelize([1]),
+                                                      "collectWithJobGroup")
+        # maxNumConcurrentTasks() is a package private API
+        max_num_concurrent_tasks = self._spark_context._jsc.sc().maxNumConcurrentTasks()
+        self.parallelism = self._decide_parallelism(max_num_concurrent_tasks, parallelism)
+        self.user_specified_parallelism = parallelism
```
maxpumperla: this isn't used anywhere, it seems. why keep it?

hyperopt/spark.py
```diff
@@ -0,0 +1,415 @@
...
+
+        if timeout is not None and (not isinstance(timeout, numbers.Number) or timeout <= 0):
```
maxpumperla: we can do the timeout check right after the super call to fail early.

hyperopt/spark.py
```diff
@@ -0,0 +1,415 @@
...
+             catch_eval_exceptions,
+             return_argmin,
+             show_progressbar,
+             ):
+        """
+        This should not be called directly but is called via :func:`hyperopt.fmin`
+        Refer to :func:`hyperopt.fmin` for docs on each argument
+        """
+
+        assert not pass_expr_memo_ctrl, "SparkTrials does not support `pass_expr_memo_ctrl`"
```
maxpumperla: this is definitely something we should document in a follow-up PR.

unit-test.md
```diff
@@ -0,0 +1,19 @@
+Running unit tests
+==============
+
+To run the  unit tests, run the `run-tests.sh` script.  You will need to set these environment variables:
+
+- `SPARK_HOME`: your local copy of Apache Spark. Look at download_spark_dependencies.sh for more details.
```
maxpumperla: you did a really good job separating the new `spark` module from the rest (for current users). For testing I'd really like if we could have an easy way for contributors to ignore or deactivate Spark tests. Why? Most hyperopt users won't use the new Spark functionality right away and I don't want to discourage new contributions. Plus, I don't think we'll get many PRs for the `spark` module.

unit-test.md
```diff
@@ -0,0 +1,19 @@
...
+- `HYPEROPT_FMIN_SEED`: the random seed. You need to get its value from `.travis.yml`.
+
+For example:
+
+```bash
+hyperopt$ HYPEROPT_FMIN_SEED=3 SPARK_HOME=/usr/local/lib/spark-2.4.0-bin-hadoop2.7 ./run_tests.sh
```
maxpumperla: Do you have explicit recommendations which Spark version to use? If so, we should document that.

hyperopt/spark.py
jkbradley: Whoops, that looks like forgotten residue, thanks!

hyperopt/spark.py
jkbradley: It's pretty arbitrary.  We did a bit of stress testing and found this seemed fine; we got input from other teams who worried about pushing it closer to 1000; I believe Spark has a hard cap around 2000; and we want to do more stress testing in the future.

I'm up for making it a changeable config.  Do you mind if we make that a follow-up?  I started a list of such items at the end of the design doc: https://docs.google.com/document/d/1R-qMVOYxaLw3HdqMiTADQDZ5z7uInMR-x42HFCXu2-s/edit#heading=h.kqrwh3pgxu44

hyperopt/spark.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @staticmethod
+    def _write_cancellation_back(trial, e):
+        trial['state'] = base.JOB_STATE_CANCEL
+        trial['misc']['error'] = (str(type(e)), str(e))
+        trial['refresh_time'] = coarse_utcnow()
+
+    def _run_trial_async(self, trial):
+        def run_task_thread():
+            # TODO: use broadcast for `domain` object
```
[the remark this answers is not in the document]
jkbradley: We actually were not able to do this because Spark's broadcast uses cPickle instead of cloudpickle.  @hanyucui would you mind removing this TODO?

I think this will be fine in general since the user's function and space should be fairly small objects.  The main worry would be if a user included a big dataset as an object in their function, rather than reading it inside their function.  I'll add a future TODO to the design doc list for checking the pickled object size.

hyperopt/spark.py
```diff
@@ -0,0 +1,415 @@
...
+
+        task_thread = threading.Thread(target=run_task_thread)
+        task_thread.setDaemon(True)
+        task_thread.start()
+        self._task_threads.add(task_thread)
+
+    def _poll_new_tasks(self):
+        new_task_list = []
+        # Make a copy of trials by slicing
+        for trial in self.trials.trials[:]:
```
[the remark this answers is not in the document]
jkbradley: During this part of the iteration, I think the only worry is that the list would be modified by adding new trials, never modified by removing trials, so I think the shallow copy would remain valid.  But I'm on board with a deeper copy too.

jkbradley:
Thanks for the quick review!

> Most importantly, let's make sure we can easily skip (or even just activate on demand) Spark tests. From hyperopt issues I know we have e.g. quite a few Windows users. While your spark dep script works well in unix envs, I don't want to have yet another "Spark installation support group" here. :D

That makes sense and is good to know!  What's the right setup or set of docs for this?  We can make sure there are documented 1-line commands to run (a) all tests and (b) all but the Spark tests.

hyperopt/spark.py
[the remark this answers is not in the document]
hanyucui: Sure, will remove it when I push a new patch.

hyperopt/tests/test_spark.py
jkbradley: Discussed with Xiangrui offline: How about we add it as `extras_require` to make it optional?

unit-test.md
jkbradley: I'd recommend we change this to 2.4.3 (latest) and in our documentation, state which versions are tested by the CI.  If users think it's needed, we could test multiple versions.

hanyucui:
@maxpumperla Sorry for the delay -- we are swamped with other stuff. Will address your comments bit by bit as we get time.

maxpumperla:
@hanyucui no worries, let me know if I can help

mblack20:
@hanyucui any progress on this?

hanyucui:
@mblack20 I am trying to spend time on this. But it will be a while before I can address all concerns.

hanyucui:
Hi @maxpumperla, thanks to @liangz1, we finally addressed your review. Can you take another look?

maxpumperla:
@hanyucui @jkbradley thanks again, that's a really cool addition to hyperopt! I'll work on a new minor release as soon as possible.

jkbradley:
Thank you @maxpumperla ... especially for your patience with us  :P 
This will be awesome to have in hyperopt!

PR #521, opened 2019-07-25
Fixes error in progressbar when using asynchronous trials

### LINKED ISSUE #520: tqdm progressbar breaks with MongoTrials
Hello,

I noticed that the progressbar when using `MongoTrials` is almost useless as it advances at one iteration a second regardless of the actual progress and then it makes as much (pseudo)iterations as it takes to complete the task.

**MWE**:

> hyperop_mongo.py

```
from hyperopt import fmin, tpe, hp, Trials
from hyperopt.mongoexp import MongoTrials
import hyperopt_to_optim

trials = MongoTrials('mongo://localhost:1234/foo_db/jobs', exp_key='exp13')

best = fmin(hyperopt_to_optim.obj, hp.uniform('x', -2, 2), trials=trials, algo=tpe.suggest, max_evals=20)
```

> hyperopt_to_optim.py

```
import time
import math
def obj(x):
    time.sleep(3)
    return math.sin(x)
```

> output:

```
77it [01:17,  1.02s/it, best loss: -0.9947311913502885]

```
(instead of 20 iterations, also there might be another issue.. why does it take so much more than 60 seconds? )

to run you need to add the directory containing `hyperopt_to_optim.py` to `PYTHONPATH` for `hyperopt-mongo-worker`

brunodoamaral:
Fixed bug in progress bar when using asynchronous trials is used.
Fixes #520

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -199,9 +199,11 @@ class FMinIter(object):
                       ) as pbar:
                 while n_queued < N:
                     qlen = get_queue_len()
+                    total_ids = 0
                     while qlen < self.max_queue_len and n_queued < N:
                         n_to_enqueue = min(self.max_queue_len - qlen, N - n_queued)
                         new_ids = trials.new_trial_ids(n_to_enqueue)
+                        total_ids += len(new_ids)
                         self.trials.refresh()
                         if 0:
                             for d in self.trials.trials:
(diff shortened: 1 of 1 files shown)


marctorrellas:
@brunodoamaral can you update this? it seemed to be fine at some point, and I'd be happy to review/approve it

maxpumperla:
@brunodoamaral I tried to pick this up, but your change does not fit into the current code base anymore. If you find a way to make this work, please open a new PR. thank you and apologies for not getting to it in the first place, when it was still working

PR #533, opened 2019-09-03
Timeout parameter to limit time (in seconds) of the search process (fmin)

vladkar:
max_time parameter was added to fmin:
- stops searching process when allowed time is passed, works similar to max_evals parameter, but limits time of search instead of number of trials
- can work together with max_evals
- max_time=60 means that search process will be stopped after 60 seconds since its start with respect to last trial (if trial requires 10 seconds and it was started at second 59, then search process takes 69 second)
- unit tests were prepared to check new parameter
- progress bar was not changed (it works based on number of trials)

Update 11 Dec 2019:

- max_time parameter was renamed to timeout to keep consistency with recent update
- timeout parameter was added to fmin function for spark implementation as well (to keep configuration of timeout and max_evals in one place)


hyperopt/fmin.py
```diff
@@ -9,6 +9,7 @@
 import os
 import sys
 import time
+import timeit
```
marctorrellas: given you only use default_timer, what do you think about changing this to `from timeit import default_timer as timer` ?

hyperopt/fmin.py
```diff
@@ -197,9 +201,9 @@ def get_queue_len():
             with tqdm(total=N+qlen, file=orig_stdout, postfix='best loss: ?',
                       disable=not self.show_progressbar, dynamic_ncols=True,
                       ) as pbar:
-                while n_queued < N:
+                while n_queued < N and (timeit.default_timer() - self.start_time) < self.max_time:
                     qlen = get_queue_len()
-                    while qlen < self.max_queue_len and n_queued < N:
+                    while qlen < self.max_queue_len and n_queued < N and (timeit.default_timer() - self.start_time) < self.max_time:
```
marctorrellas: are u sure u need to check the time twice? can line 205 take long to run?

hyperopt/tests/test_fmin.py
```diff
@@ -212,3 +214,36 @@ def test_generate_trials_to_calculate(self):
                     )
         assert best['x'] == 0.0
         assert best['y'] == 0.0
+
+
+def test_max_time():
```
marctorrellas: this probably can be written in a for loop. Ideally with pytest it would be with pytest.mark.parametrize but hyperopt doesn't use pytest yet

marctorrellas:
see https://github.com/hyperopt/hyperopt/pull/311. It seems to me the aim is the same, so it'd be nice to work together and close 2 PRs in 1 effort, if possible

hyperopt/fmin.py
vladkar: Agree

hyperopt/fmin.py
vladkar: It is not required, condition was optimized

vladkar:
Pull request was updated:
 - max_time parameter was renamed to timeout to keep consistency with recent update
 - timeout parameter was added to fmin function (to keep configuration of timeout and max_evals in one place)

There are some failed tests at spark functionality, but I got the same failed tests even before my changes (from clean master branch)

hyperopt/fmin.py
```diff
@@ -234,7 +239,8 @@ def get_n_unfinished():
         ) as progress_ctx:
 
             all_trials_complete = False
-            while n_queued < N or (block_until_done and not all_trials_complete):
+            while (n_queued < N or (block_until_done and not all_trials_complete)) and \
+                    (self.timeout is None or (timer() - self.start_time) < self.timeout):
```
marctorrellas: this can be  `timer() - self.start_time < (self.timeout or float('inf'))`. Also, please use black to format code

hyperopt/base.py
```diff
@@ -226,6 +228,18 @@ def spec_from_misc(misc):
     return spec
 
 
+def validate_timeout(timeout):
+    if timeout is not None and (
+            not isinstance(timeout, numbers.Number)
+            or timeout <= 0
+            or isinstance(timeout, bool)
```
marctorrellas: I think this is redundant. If it's not a number, you don't need to check if it's a bool

marctorrellas: also, if the docstring says int, then I think better to check here is int, rather than any number

hyperopt/fmin.py
```diff
@@ -371,6 +378,10 @@ def fmin(
     max_evals : int
         Allow up to this many function evaluations before returning.
 
+    timeout : int
```
marctorrellas: please use style from other params :  

```
timeout: None or int.
Limits ...
If None, then the search process has no time constraint.
```

hyperopt/tests/test_fmin.py
```diff
@@ -235,3 +237,36 @@ def test_generate_trials_to_calculate(self):
         )
         assert best["x"] == 0.0
         assert best["y"] == 0.0
+
+
+def test_timeout():
+    fn = lambda x: [time.sleep(1), x][1]
+    space = hp.choice('x', range(20))
```
marctorrellas: please use black

vladkar:
Ok, stuff was resolved

hyperopt/tests/test_fmin.py
vladkar: True, it could be a cycle, but I intentionally keep it like this to make test readable and fast (taking into account that we literally need to wait to test timeout) as possible.

hyperopt/base.py
vladkar: This is not my code segment, I used check from spark implementation, but since I use the same parameter I just moved this check to separate method to base file. To not break something by accident I did not change the logic of this validation.

marctorrellas: 1) if you can reuse code from somwhere else, do it. Do not duplicate
2) Either the docstring or check need to be changed, to match. The check allows for floats and has a check for bool that seems useless to me, while the docstring states int. I would change this code as I don't think we need float precision for a timeout in seconds

hyperopt/fmin.py
vladkar: It can, but version I used (a) subjectively more readable, (b) objectively faster

hyperopt/tests/test_fmin.py
vladkar: Done

hyperopt/fmin.py
vladkar: Black was applied

hyperopt/fmin.py
vladkar: Done

hyperopt/base.py
vladkar: 1. It was reused, not duplicated. As I mention, this check was just moved to base.
2. I think this should be discussed when this check was introduced to spark implementation. Again, I prefer to reuse it than make (almost) the same new check.

vladkar:
I guess there is issue with test 'test_task_maxFailures_warning', its result is inconsistent (sometimes it is ok, sometimes it is failed) even on pure master branch.

marctorrellas:
I don't know about that test. Could you please make it deterministic?

marctorrellas:
> I don't know about that test. Could you please make it deterministic?

forget about it, seems to work. Ok, will review when I can and approve. @maxpumperla  can you take a look at this?

maxpumperla:
@marctorrellas we need to look at unrelated tests separately

@vladkar great work, I have nothing to add. moving that check to base is fair, you don't have to clean up other PRs here. :+1: 

anything else you wanted to change or are we good to go from your side?

vladkar:
@marctorrellas @maxpumperla thank you for youк feedback, that is it from my side for this PR.

marctorrellas:
@vladkar good job, thanks for your contribution

marctorrellas:
@maxpumperla when is the next release expected? this and the functionality about verbosity are cool things I would like see soon :)   count on me if you need a hand

maxpumperla:
@marctorrellas I want to fix that (apparently still around) BSON issue (https://github.com/hyperopt/hyperopt/issues/547) and look into lower/upper bounds for randint (https://github.com/hyperopt/hyperopt/issues/536) as a minimum here. got time to look at any one of these?

PR #535, opened 2019-09-27
Convert output of fmin to same dtypes in space

jmmonteiro:
When training a Gradient Boosting Regression model (from sklearn) on training data like this:

```
space = {
        "max_depth": scope.int(hp.quniform("max_depth", 1, 10, 1)),
        "n_estimators": scope.int(hp.quniform("n_estimators", 1, 111, 10)),
        "subsample": hp.uniform("subsample", 0.01, 1),
    }
    f_min_obj = partial(cv_wrapper, X_input=X_tr, y_input=y_tr)
    best = fmin(fn=f_min_obj, space=space, algo=tpe.suggest, max_evals=50)
```

The dictionary returned by `fmin` will have all its parameters as floats, if I try to train the model with the best parameters using: `model = GradientBoostingRegressor(**best).fit(X_tr, y_tr)`, it crashes since `max_depth` and `n_estimators` should be `int`.

The solution is a bit hacky, since I'm declaring `out_dict`, however, I was not able to change `trials.argmin` on my machine. Any comments on how to make this better are more than welcome.

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -423,7 +423,16 @@ def fmin(fn, space, algo, max_evals, trials=None, rstate=None,
     if return_argmin:
         if len(trials.trials) == 0:
             raise Exception("There are no evaluation tasks, cannot return argmin of task losses.")
-        return trials.argmin
+        # Make sure the arguments are returned with the correct type
+        out_dict = dict()
+        for arg_idx, arg_val in trials.argmin.items():
+            try:
+                if space[arg_idx].name in ["int", "float"]:
+                    arg_val = eval(space[arg_idx].name + "(arg_val)")
+            except KeyError:
+                print(arg_idx + " not found in space.")
+            out_dict[arg_idx] = arg_val
+        return out_dict
     elif len(trials) > 0:
         # Only if there are some succesfull trail runs, return the best point in the evaluation space
         return space_eval(space, trials.argmin)


jmmonteiro:
Just noticed that this solution breaks a few unit tests. I'll have a further look into how to do this type conversion safely.

hyperopt/fmin.py
```diff
@@ -424,6 +424,16 @@ def fmin(fn, space, algo, max_evals, trials=None, rstate=None,
         if len(trials.trials) == 0:
             raise Exception("There are no evaluation tasks, cannot return argmin of task losses.")
         return trials.argmin
+        # Make sure the arguments are returned with the correct type
```
marctorrellas: I'm not sure this is doing anything if it goes after a return statement

jmmonteiro: You're right. Thank you for spotting that.
Probably a bad fix of merge conflicts. I'll fix it as soon as I have a bit of time.

jmmonteiro: Fixed.
Squashed both commits into a single commit.

marctorrellas:
Even though I'm not totally against of returning the correct types, hyperopt has its own mechanism for this:

```
best = fmin(...)
model = GradientBoostingRegressor(**space_eval(space, best))
```

This is explain at the end of the docstring of `fmin`:

```
Returns
    -------

    argmin : dictionary
        If return_argmin is True returns `trials.argmin` which is a dictionary.  Otherwise
        this function  returns the result of `hyperopt.space_eval(space, trails.argmin)` if there
        were succesfull trails. This object shares the same structure as the space passed.
        If there were no succesfull trails, it returns None.```

jmmonteiro:
Thanks for the feedback @marctorrellas 
I guess that using `return_argmin=False` would make this pull request redundant. Happy to still work on it if you think it's worth it, otherwise I'll just close it.

marctorrellas:
Exactly, you can do either way: with space_eval or with return argmin

My feeling is that the PR objective can be achieved using what hyperopt provides. However, I think it's confusing both the documentation and name of the variable so if you want to work on a refactoring to make it clearer, I would be happy to review (please change title of PR in such a case)

marctorrellas:
@jmmonteiro please update the PR or close it

maxpumperla:
@marctorrellas thanks for chiming in here. I think `space_eval` is what we want here. I think the best solution would be to close this PR for now and instead invest some time in a great suite of advanced examples that show proper usage. Our docs still need a lot of improvement. @jmmonteiro I hope that's OK for you, I don't want to discourage contributions...

jmmonteiro:
Hi @maxpumperla 
It's perfectly fine, makes perfect sense. I wouldn't have opened the pull request if I knew about `space_eval`. Thanks for your work on hyperopt!

maxpumperla:
@jmmonteiro thanks for understanding, and you're welcome! I'll try to get in a "docs pass" to add more FAQs like this, because this certainly comes up here and there. :+1:

Issue #353, opened 2018-01-17
Separate mongodb dependency


jaberg:
@aldenor wrote (#350): 

I understand that Mongo backend is an essential part of hyperopt, however quite often it won't be needed (people just import fmin and hp and go on optimizing). It would be nice if, at the very least, the core code didn't try to import pymongo unless you tried to create a MongoTrials.

It might be nice to provide a clean separation, so that mongo-related code is not intertwined with the core codebase; also, pymongo could be made a separate dependency, enabled via [mongo] setuptools feature (which would also install the mongo backend), like so:

```
$ pip install hyperopt[mongo]
```

ClimbsRocks:
Whoa, I didn't realize you could specify dependencies that way. Do you have any favorite examples of this? 

I'm also a big fan of packages that just work out of the box (don't require db installs or db service starts or anything else), that allow the user to opt into more advanced features when they want, and choose to take on that incremental setup work (ie, the mongo features). 

Really glad to see this project coming back to life!

palicand:
Hello. I have created a PR that should fix this. If there anything else related to the issue is needed, I'll be happy to add it there.

PR #537, opened 2019-10-03
Make mongo an extra dependency

palicand:
`pymongo` should be a dependency only for those, who want to use MongoTrials.

Note that I also had to change how `hyperopt` is installed in the CI:

* `pip install .[MongoTrials]` because we want to specify installation of `pymongo`, so that tests for `mongoexp` run correctly
* `pip install pytest>=3.6 pytest-cov pep8 pytest-pep8` was causing CI errors for Python 2.7 and 3.6, so I split it into two separate installs - first `pytest` and then the rest.

See also #353

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ services:
   - mongodb
 
 before_script:
-  - sleep 15  # mongo takes time to start
+  - sleep 15 # mongo takes time to start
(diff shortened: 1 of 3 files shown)


leftys:
Looking forward to see this merged!

maxpumperla:
@palicand great addition, thank you!

PR #546, opened 2019-10-18
Change ModuleNotFoundError to ImportError to work in Python < 3.6

BartoszCki:
There is no ModuleNotFoundError in Python2. This caused NameError in py<3

ModuleNotFoundError inherits from ImportError so changing the exception should work for both Py2 and Py3

### DIFF
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -12,7 +12,7 @@ from hyperopt.utils import coarse_utcnow, _get_logger, _get_random_id
 try:
     from pyspark.sql import SparkSession
     _have_spark = True
-except ModuleNotFoundError as e:
+except ImportError as e:
     _have_spark = False
 
 logger = _get_logger('hyperopt-spark')


BartoszCki:
Just found out this affects all Python version < 3.6 so it fails on official tensorflow docker images - they are (God knows why) made with Python3.5

marctorrellas:
The last milestone was

image

I would say there's not much intention to continue with Python2 support...

BartoszCki:
It doesn't work with Python3.5 too

marctorrellas:
Python 3.6 was released on Dec 2016, maybe it's a bare minimum to have this version? (considering there's already 3.7 and 3.8). In other words, do you think the PR could be declined? what's your interest in getting this in?

BartoszCki:
It's End Of Life is set to be October 2020 so it's still a lot of time.
Official Tensorflow Docker images still run on Python3.5

maxpumperla:
@BartoszCki sorry for not getting around earlier, good catch. I'll cut out `0.2.2` now to patch this.

Issue #547, opened 2019-10-18
Line 399: Error message: "module 'bson' has no attribute 'BSON'"


rossmpowell:
Hi all, 

I am working on a machine learning project, and using the Hyperas package to tune my model. Hyperas uses two functions to setup a tuning job: a data function, and model function. So far I've had some issues with them talking to each other during the encoding stage of building the model because the bson.BSON object called in hyperopt/base.py does not exist.

I ended up checking out the bson code found here: https://github.com/py-bson/bson, and the bson.BSON object does not exist there either. Is there a different bson distribution or package that I should be using?
____________________________________________________________________________________
Notes:
- I'm using an AWS SageMaker notebook, that could also have something to do with bson being imported weirdly behind the scenes.
- A somewhat related issue: https://github.com/maxpumperla/hyperas/issues/23

Thanks!

rossmpowell:
Update (Solution):

The sequence of imports below worked for getting bson a moment ago

pip install bson 
pip install hyperopt
pip install hyperas
sudo pip uninstall bson
pip install pymongo

(Solution is similar to what was referenced in maxpumperla/hyperas#23)

PhilipMay:
I have the same problem (see below). Although there seems to be a workaround: can someone please fix this?

```
[...]
  File "/home/phmay/miniconda3/envs/tf2/lib/python3.6/site-packages/mltb/hyperopt.py", line 99, in fmin
    show_progressbar=show_progressbar,
  File "/home/phmay/miniconda3/envs/tf2/lib/python3.6/site-packages/hyperopt/fmin.py", line 403, in fmin
    show_progressbar=show_progressbar,
  File "/home/phmay/miniconda3/envs/tf2/lib/python3.6/site-packages/hyperopt/base.py", line 651, in fmin
    show_progressbar=show_progressbar)
  File "/home/phmay/miniconda3/envs/tf2/lib/python3.6/site-packages/hyperopt/fmin.py", line 422, in fmin
    rval.exhaust()
  File "/home/phmay/miniconda3/envs/tf2/lib/python3.6/site-packages/hyperopt/fmin.py", line 276, in exhaust
    self.run(self.max_evals - n_done, block_until_done=self.asynchronous)
  File "/home/phmay/miniconda3/envs/tf2/lib/python3.6/site-packages/hyperopt/fmin.py", line 225, in run
    self.trials.insert_trial_docs(new_trials)
[...]

Issue #541, opened 2019-10-16
cannot import name 'BSON' on new release 0.2


grfreitas:
When calling fmin:

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<command-1644382932079013> in <module>()
----> 1 from bson import BSON

ImportError: cannot import name 'BSON'
```

richardliaw:
+1

Chriskuei:
+1

WeiwenRen:
+1

kiran-kate:
Looks like commit 4dae46c4724d4796c6a83f146e8f336013ad9bdf caused this change. Any plans to revert it? Otherwise, packages dependent on Hyperopt are locked in with version 0.1.2 or have to handle the bson dependency themselves.

Spinachboul:
if pymongo has its own internal bson package, then how are we supposed to use bson in our code, and how are we supposed to import it?

PR #548, opened 2019-10-21
Change the import of SON

s6juncheng:
from bson import SON cause import error

PR #549, opened 2019-10-22
Fix BSON import issue

palicand:
A package `bson` has been added to the `setup.py` in commit 4dae46c4724d4796c6a83f146e8f336013ad9bdf,
because `pymongo` has been removed. The package is however not compatible
with the `bson` package from `pymongo` and has been causing
`ImportError`s.

Therefore we add all the imports of `bson` to a guarded block.
It should be safe to do so, because any code that needs the `bson`
package is also dependent on `pymongo`.

Issues: #541 and #547

### DIFF
--- a/hyperopt/mongoexp.py
+++ b/hyperopt/mongoexp.py
@@ -110,11 +110,11 @@ import numpy
 try:
     import pymongo
     import gridfs
+    from bson import SON
     _has_mongo = True
 except:
     _has_mongo = False
 
-from bson import SON
 from .base import JOB_STATES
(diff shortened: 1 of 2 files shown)


marctorrellas:
it looks to me like this should be combined with #548

marctorrellas:
also, I was getting problems with BSON so had to change 

bson.BSON.encode(trial) —> bson.BSONCoding.bson_encode(trial)

@maxpumperla can you take a look at this? 🙏

maxpumperla:
@palicand thanks for the fix and @marctorrellas thanks for pinging me. :D

PR #556, opened 2019-10-30
hyperopt facelift with black

marctorrellas:
This is a very straightforward PR to start using `black` to format the code

Main contribution is to apply black to the whole code

Update: Initially the idea was to force to use black but as it requires python 3.6+, all the following details in this description have been removed / postponed

Plus, I've added instructions in the README.md for contributors. Basically it's as simple as running

pre-commit install

once you clone. Then, every commit will run `black` on the modified files and fail if there are any files that needed changes. There are two options here:

* To run black and if there are files that need changes fail and do nothing. This means if you commit again the commit won't succeed, but it doesn't change your files.
* To run black and if there are files that need changes fail and modify them. This means if you commit again the commit would succeed. Good thing is you don't need to run black in every file

I chose the latter. Let me know your thoughts about this

About the tests: it passes the tests for python3.6, but not for the other 2 distros. This is because black requires 3.6+

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,13 +60,7 @@ git clone https://github.com/jaberg/hyperopt.git
 cd hyperopt && python setup.py develop &&  pip install -e '.[MongoTrials, SparkTrials, ATPE]'
 ```
 
-Then install the pre-commit hook as
-```
-pre-commit install
-```
-
-so that black is automatically run (and failing if required) every time you 
-commit
+We recommend to use `black` to format your code before submitting a PR
 
 ## Algorithms
 
--- a/README.md
+++ b/README.md
@@ -66,7 +66,10 @@ Currently three algorithms are implemented in hyperopt:
 
 Hyperopt has been designed to accommodate Bayesian optimization algorithms based on Gaussian processes and regression trees, but these are not currently implemented.
 
-All algorithms can be run either serially, or in parallel by communicating via [MongoDB](https://mongodb.com).
+All algorithms can be parallelized in two ways, using:
+
+- [Apache Spark](https://spark.apache.org/)
+- [MongoDB](https://mongodb.com)
 
 ## Documentation
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ Hyperopt is a Python library for serial and parallel optimization over awkward
 search spaces, which may include real-valued, discrete, and conditional
 dimensions.
 
+## Requirements
+
+python 3.6+
+
 ## Getting started
 
 Install hyperopt from PyPI
@@ -40,12 +44,12 @@ space = hp.choice('a',
     ])
 
 # minimize the objective over the space
-from hyperopt import fmin, tpe
+from hyperopt import fmin, tpe, space_eval
 best = fmin(objective, space, algo=tpe.suggest, max_evals=100)
 
-print best
+print(best)
 # -> {'a': 1, 'c2': 0.01420615366247227}
-print hyperopt.space_eval(space, best)
+print(space_eval(space, best))
 # -> ('case 2', 0.01420615366247227}
 ```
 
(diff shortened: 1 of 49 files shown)


README.md
```diff
@@ -12,6 +12,10 @@ Hyperopt is a Python library for serial and parallel optimization over awkward
 search spaces, which may include real-valued, discrete, and conditional
 dimensions.
 
+## Requirements
+
+python 3.6+
```
maxpumperla: @marctorrellas I missed the part about req. 3.6+. As you know I'm willing to drop Py 2 soon, but I'm not convinced it's wise to exclude 3.4 etc.

Good to see that applying this wouldn't be too much work, all in all

marctorrellas: to be honest, I didn't know either, but black requires 3.6. what we can do as a preliminary step is just to blacken the codebase, without changing the requirements or hooks. This will at least clean the actual state and hopefully it will be less effort in the future

marctorrellas: what do you think @maxpumperla ? otherwise I'll close this

maxpumperla: I find this too restrictive for now. let's come back to this once we're ready with dropping python2 support. one step at a time :D

marctorrellas: oh, forgot to remove this in README. Now it's just black applied, no constraints on Python or hooks

maxpumperla:
@marctorrellas can you please remove the "python 3.6+" part from the readme? it's misleading. Instead, I think we should add new dependencies in `setup.py`, namely:

```python
        "dev": ["black", "pre-commit"],
```

the README should then read as follows:

```shell script
cd hyperopt && python setup.py develop &&  pip install -e '.[MongoTrials, SparkTrials, ATPE, dev]'
```
and I think we should definitely add a quick section that tells developers how to leverage automatic black linting by first running `pre-commit install` and add a small note there that `dev` dependencies and black use python 3 under the hood.

marctorrellas:
> @marctorrellas can you please remove the "python 3.6+" part from the readme? it's misleading. Instead, I think we should add new dependencies in `setup.py`, namely:
> 
> ```python
>         "dev": ["black", "pre-commit"],
> ```
> 
> the README should then read as follows:
> 
> ```shell
> cd hyperopt && python setup.py develop &&  pip install -e '.[MongoTrials, SparkTrials, ATPE, dev]'
> ```
> 
> and I think we should definitely add a quick section that tells developers how to leverage automatic black linting by first running `pre-commit install` and add a small note there that `dev` dependencies and black use python 3 under the hood.

@maxpumperla the only thing in the README is this now:

We recommend to use `black` to format your code before submitting a PR

I'm happy to add your changes though

marctorrellas:
> > @marctorrellas can you please remove the "python 3.6+" part from the readme? it's misleading. Instead, I think we should add new dependencies in `setup.py`, namely:
> > ```python
> >         "dev": ["black", "pre-commit"],
> > ```
> > 
> > 
> > the README should then read as follows:
> > ```shell
> > cd hyperopt && python setup.py develop &&  pip install -e '.[MongoTrials, SparkTrials, ATPE, dev]'
> > ```
> > 
> > 
> > and I think we should definitely add a quick section that tells developers how to leverage automatic black linting by first running `pre-commit install` and add a small note there that `dev` dependencies and black use python 3 under the hood.
> 
> @maxpumperla the only thing in the README is this now:
> 
> We recommend to use `black` to format your code before submitting a PR
> 
> I'm happy to add your changes though

done

README.md
```diff
@@ -40,22 +40,35 @@ space = hp.choice('a',
...
+Note that dev dependencies require black, using python 3.6+ under the hood.
+
+We recommend to use `black` to format your code before submitting a PR. You can use it 
+with a pre-commit hook as follows:
+
+```
+pre-commit install
+```
+
+Then, once you commit ensure than git hooks are activated (Pycharm for example have the
```
maxpumperla: typo: than --> that
typo: Pycharm have --> has

marctorrellas: done

maxpumperla:
@marctorrellas great, thank you! looking forward to more manifestations of your ideas. :+1:

Issue #553, opened 2019-10-25
It seems that atpe does not support int or choice!


hfarazi:
It seems that atpe does not support int or choice! Is it a bug or limitation of the method?

FraserGrid:
I feel as though this may be related to my issue too... https://github.com/hyperopt/hyperopt/issues/552

denizetkar:
I suffer from this problem as well. Actually I was trying to speed up evaluation of previously evaluated points. Then I came up with the following code snippet (which doesn't work because 'atpe' does not support 'choice'):

```
import hyperopt
from hyperopt import hp
from hyperopt.pyll.stochastic import sample
from hyperopt.fmin import generate_trials_to_calculate

space = {'a': hp.uniform('a', 0, 1), 'b': hp.normal('b', 0, 1), 'loss': hp.pchoice('loss_exists', [(1, None), (0, hp.normal('loss', 0, 1))])}

def objective(params):
    if params['loss'] is not None:
        return params['loss']
    return params['a'] + params['b']

trials = generate_trials_to_calculate([{'a': 0, 'b': -3, 'loss_exists': 1, 'loss': -3}])
best_params = hyperopt.fmin(objective, space, algo=hyperopt.atpe.suggest, max_evals=1, trials=trials)
print(best_params)
```

Basically, I am looking for a way to kickstart the optimization with a list of points to evaluate and I also happen to have their loss values which I want to pass to the objective function. How can I do that while using ATPE?
(NOTE: all of my other hyper parameters are compatible with ATPE, just like the `'a'` and `'b'` in the example above)

genixpro:
ATPE should support int & choice, let me investigate

genixpro:
Have you gotten any errors or stacktraces when you attempt to use these hyperparameters?

jithurjacob:
Any updates on this. I'm using 0.2.3 and is facing issues with pchoice and atpe

BJWiley233:
So atpe is working with hp.choice but unfortunately for hp.uniform you get an index error in the recursion loop at `Hyperparameter.createHyperparameterConfigForHyperoptDomain` [here]()

works for choice eventually `fmin` gets here
```
ffn_hidden = np.array([150, 200, 250, 300, 400])
search_space = {
    'ffn_hidden': hp.choice('ffn_hidden', ffn_hidden),
}
## eventually fmin gets to something like
domain=search_space['ffn_hidden']
[...]

PR #557, opened 2019-10-30
Fixed an issue where the ATPE algorithm fails with randint and choice parameters

genixpro:
Partial fix for issue #553


maxpumperla:
thanks, Brad!

PR #559, opened 2019-10-31
Fixes tqdm progressbar for parallel execution of fmin.

samhiatt:
tqdm progress bar was incrementing after each poll interval, regardless of whether any trials had finished or not. This change only increments the progress bar when a trial has completed and waits until all trials have returned before exiting (if `block_until_done` is set).

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -201,13 +201,25 @@ class FMinIter(object):
         def get_queue_len():
             return self.trials.count_by_state_unsynced(base.JOB_STATE_NEW)
 
+        def get_n_done():
+            return self.trials.count_by_state_unsynced(base.JOB_STATE_DONE)
+
+        def get_n_unfinished():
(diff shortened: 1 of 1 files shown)


maxpumperla:
@samhiatt awesome, thank you!

jkbradley:
I was just looking into this issue but was happy to see this patch!  I tested it and think it works for regular Trials, but I think it completely removes the progress bar for SparkTrials.  (I'm not sure about MongoTrials yet.)  Did you look into how to get it to work with all 3 types of Trials?

maxpumperla:
@jkbradley yeah, that's unfortunate. but the progress bar was broken for such a long time now that I was happy to see that fix. better to be on the conservative side here

samhiatt:
I haven't yet had a chance to look at the spark code to see how progress bar updating can be implemented.  
If `trials.count_by_state_unsynced([base.JOB_STATE_NEW, base.JOB_STATE_RUNNING])` and `trials.count_by_state_unsynced(base.JOB_STATE_DONE)` return appropriate counts then it should work.

PR #560, opened 2019-10-31
Calls trials.refresh() before reporting best loss in progress bar

samhiatt:
Now immediately shows the result of the first trial after it completes, and keeps the best loss up to date. (Before it was lagging behind by one trial)


maxpumperla:
@samhiatt perfect, thanks. just reran CI for 3.5 and seems to be good to go

PR #563, opened 2019-11-04
typo when making losses positive

nfultz:
When minval is negative, it needs to be subtracted out in order to make the result positive. Adding it will make the loss more negative.


maxpumperla:
@nfultz great catch!

PR #570, opened 2019-11-18
Warn user about spark task retries in SparkTrials

jkbradley:
Spark will, by default, retry tasks 4 times upon failure; see `spark.task.maxFailures` in https://spark.apache.org/docs/latest/configuration.html

With SparkTrials, it is unclear if users will really want that: A failed trial might be a fluke (in which case retries are nice), or it might indicate bad hyperparameters (e.g., causing an OOM from the model being too large).  This issue can be bad when the trials take a long time since retries can quadruple the time before the user sees an issue.

This patch adds a warning to the user informing them about this issue.  The warning is only printed once per fmin() call, and it is only printed if no trials return for a full minute.

### DIFF
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -73,8 +73,9 @@ class SparkTrials(Trials):
                                     isinstance(timeout, bool)):
             raise Exception("The timeout argument should be None or a positive value. "
                             "Given value: {timeout}".format(timeout=timeout))
-        self._spark_context = SparkSession.builder.getOrCreate().sparkContext if spark_session is None \
-                                  else spark_session.sparkContext
+        self._spark = SparkSession.builder.getOrCreate() if spark_session is None \
+                      else spark_session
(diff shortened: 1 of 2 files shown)


hyperopt/spark.py
```diff
@@ -73,8 +73,9 @@ def __init__(self, parallelism=None, timeout=None, spark_session=None):
                                     isinstance(timeout, bool)):
             raise Exception("The timeout argument should be None or a positive value. "
                             "Given value: {timeout}".format(timeout=timeout))
-        self._spark_context = SparkSession.builder.getOrCreate().sparkContext if spark_session is None \
-                                  else spark_session.sparkContext
+        self._spark = SparkSession.builder.getOrCreate() if spark_session is None \
```
jkbradley: This handle to a SparkSession is useful since SparkSession is a more reliable way (than a SparkContext) to get a conf value.

PR #577, opened 2019-11-26
Fix tqdm error on close call

tonypizzicato:
When hyperopt fails with exception and python logging is used the following error occurs:

```
Error in atexit._run_exitfuncs:                                                                                            
Traceback (most recent call last):                       
  File "/usr/lib/python3.7/logging/__init__.py", line 2039, in shutdown     
    h.close()                                                                                                                       
  File "/home/tony/.local/lib/python3.7/site-packages/absl/logging/__init__.py", line 854, in close
    self.stream.close()                                                                                                
AttributeError: 'DummyTqdmFile' object has no attribute 'close'      
```


maxpumperla:
@tonypizzicato thank you!

PR #581, opened 2019-12-01
Fix verbosity for fmin with tpe.suggest

marctorrellas:
fmin used to have an unused argument `verbose`. With this PR all the messages can be removed with verbose=False. Doing this also removes the progressbar.
I also introduced a verbose argument for `tpe.suggest`

As a byproduct of this PR:
- Some pieces of code in `if 0` were removed
- Some strings exceeding max line length were split

### DIFF
--- a/hyperopt/base.py
+++ b/hyperopt/base.py
@@ -684,7 +684,7 @@ class Ctrl(object):
     """
 
     info = logger.info
-    warn = logger.warn
+    warn = logger.warning
     error = logger.error
     debug = logger.debug
 
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -26,7 +26,8 @@ try:
     import cloudpickle as pickler
 except Exception as e:
     logger.info(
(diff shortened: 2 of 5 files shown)


marctorrellas:
@maxpumperla  `pip install .[SparkTrials]` is failing for 2.7. Any idea? Tests are passing for 3.5 and 3.6

maxpumperla:
@marctorrellas thanks. no worries, travis is sometimes a bit flaky.

marctorsoc:
Do you mean that the `verbose` param in `fmin` is not enough to control the verbosity? Can you show a minimum example program and what's your issue?

mtwhitecamel:
@skoeb 's problem is algo(tpe.suggest by default) use a param named verbose, which controlling several logging function,
but when fmin invoke algo() in line 278: new_trial = algo(new_ids, self.domain, trials, self.rstate.integers(2 ** 31 - 1)),
it doesn't pass self.verbose to tpe.suggest,
maybe change to new_trial = algo(new_ids, self.domain, trials, self.rstate.integers(2 ** 31 - 1), verbose=self.verbose) is more reasonable

Issue #476, opened 2019-02-28
tqdm progress bars break when used inside objectives


nalourie-ai2:
When an objective function uses a tqdm progress bar, the output is not properly removed from the screen. The bug occurs both when `fmin` is used with `show_progressbar=False` as well as when it's used with `show_progressbar=True`. 

I obtained this issue using hyperopt version `0.1.2`. The issue doesn't occur in version `0.1.1`.

Here is my python environment:

```
decorator==4.3.2
future==0.17.1
hyperopt==0.1.2
networkx==2.2
numpy==1.16.2
pymongo==3.7.2
scipy==1.2.1
six==1.12.0
tqdm==4.31.1
```

And here is a minimal example to reproduce the issue:

```python
import time

from hyperopt import fmin, hp, tpe
import tqdm

def objective(x):
    for _ in tqdm.tqdm(range(2)):
        time.sleep(0.5)
    return x


print('\n\n#### Bug with ``show_progressbar=True`` ####\n\n')

fmin(
    fn=objective,
    space=hp.uniform('x', 0, 1),
    algo=tpe.suggest,
    max_evals=3,
    show_progressbar=True)


print('\n\n#### Bug with ``show_progressbar=False`` ####\n\n')

fmin(
    fn=objective,
    space=hp.uniform('x', 0, 1),
    algo=tpe.suggest,
    max_evals=3,
    show_progressbar=False)
```

I believe this issue is caused by `std_out_err_redirect_tqdm` context manager, though I have not investigated enough to verify whether or not that's true.

PhilipMay:
Ohh that's bad. Sorry. I will investigate this. I hope you found a workaround.

maxpumperla:
@PhilipMay that was precisely the question I had for you before merging that PR. It's a little unfortunate, but not that big of a deal. If you can please provide a fix relatively soon, that'd be great. Otherwise I'll have to revert the PR and cut a new minor release.

PhilipMay:
Well. Redirect writing (https://github.com/tqdm/tqdm#redirecting-writing) combined with nested progress bars (https://github.com/tqdm/tqdm#nested-progress-bars) does not seem to work. Either you do not redirect and break the progressbar when you do any `print()` output or you redirect and that breaks the nesting. I can not provide a fix or improvement for this. I am sorry. :-(

david-waterworth:
[...]

PR #583, opened 2019-12-03
Make tqdm progress optional

chrisavl:
This change makes the progress reporting customizable, and makes it possible to avoid the call to `std_out_err_redirect_tqdm`.

I believe it can also solve the problem described in https://github.com/hyperopt/hyperopt/issues/476#issuecomment-560026977

I kept the show_progressbar argument to be backwards compat, but I'm not sure it is necessary?

Due to the indent change, the diff might look strange on GH. `git diff master..progress-callback -w -- hyperopt/fmin.py` is a bit easier to read.

### DIFF
--- a/hyperopt/tests/test_progress.py
+++ b/hyperopt/tests/test_progress.py
@@ -1,6 +1,4 @@
-from __future__ import print_function
-from builtins import range
-from unittest.mock import patch
+from future import standard_library
 import sys
 import io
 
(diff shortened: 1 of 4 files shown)


maxpumperla:
@chrisavl awesome, thank you

PhilipMay:
@chrisavl many thanks :-)

@maxpumperla can you please prepare a release soon?

hyperopt/tests/test_progress.py
```diff
@@ -0,0 +1,14 @@
+from future import standard_library
+import sys
+import io
+
+from hyperopt.progress import tqdm_progress_callback, no_progress_callback
```
easyas314159: You're missing a unit test for `no_progress_callback `

marctorrellas:
> @chrisavl many thanks :-)
> 
> @maxpumperla can you please prepare a release soon?

@PhilipMay see https://github.com/hyperopt/hyperopt/pull/533 for the blockers for the new release. I'm not sure the former is a problem anymore, as I commented here https://github.com/hyperopt/hyperopt/issues/547

PR #588, opened 2019-12-17
Improving tests

marctorrellas:
This PR improves tests a bit by removing a try/except in `test_rdists`. This was failing because the new versions of scipy require pytest for the submodules imported

I also added [nose, pytest] to the dev list, so that devs get them when following the README instructions. Let me know if you have a better solution for this

As a byproduct I refactored some of the code and fixed one seed for fmin in `test_domains` so that tests should be totally/more deterministic

Finally, added a warning as one of the tests was skipped. I don't know how to fix it, so at least this will be there and someone will eventually work on it. Happy to get advice on this

### DIFF
--- a/hyperopt/rdists.py
+++ b/hyperopt/rdists.py
@@ -10,6 +10,7 @@ import numpy as np
 import numpy.random as mtrand
 import scipy.stats
 from scipy.stats import rv_continuous  # , rv_discrete
+from scipy.stats._continuous_distns import lognorm_gen as scipy_lognorm_gen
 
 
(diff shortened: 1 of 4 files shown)


maxpumperla:
@marctorrellas thank you! I didn't know about this test either tbh, good catch. We'll have to revisit this at some point.

marctorrellas:
I would like this to be in pytest, but that's a lot of work I personally don't want to do :D

maxpumperla:
@marctorrellas ah, right. I remember you wanted to push for pytest ;)

Issue #267, opened 2016-09-07
Documentation for saving and reloading evaluations with Trials


bjkomer:
Currently the wiki is not very clear that it is possible to save a set of evaluations and then continue where they were left off using the `Trials` object. It would be nice if a small example was added to the wiki that shows how to do this and mentions that the `max_evals` parameter refers to the total number of items in the trials database, rather than the number of evals to do for that specific `fmin` call.

There doesn't seem to be a nice way to do a pull request on a wiki, but I'm envisioning something like this:

```
import pickle
from hyperopt import fmin, tpe, hp, Trials, STATUS_OK

def objective(x):
    return {'loss': x ** 2, 'status': STATUS_OK }

# Initialize an empty trials database
trials = Trials()

# Perform 100 evaluations on the search space
best = fmin(objective,
    space=hp.uniform('x', -10, 10),
    algo=tpe.suggest,
    trials=trials,
    max_evals=100)

# The trials database now contains 100 entries, it can be saved/reloaded with pickle or another method
pickle.dump(trials, open("myfile.p", "wb"))
trials = pickle.load(open("myfile.p", "rb"))

# Perform an additional 100 evaluations
# Note that max_evals is set to 200 because 100 entries already exist in the database
best = fmin(objective,
    space=hp.uniform('x', -10, 10),
    algo=tpe.suggest,
    trials=trials,
    max_evals=200)

print(best)
```

baziotis:
@bjkomer Very nice! I was looking for something like this. I have one question.

When we load a saved trials object and pass it to the `fmin`, does hyperopt take into account the previous runs? This is important in the case `algo=tpe.suggest`. Have you verified that?

bjkomer:
@baziotis Good question! Yes, it does take previous runs into account. 

I just confirmed this on the toy example above by setting the `rstate` parameter in both the `fmin` calls, and then running it with and without the first `fmin` call, the final result is the exact same. In other words, whether you do two sets of 100 evaluations or one set of 200 evalutations, the result is the same.

I also did some digging in the code to be sure. [Line 156](https://github.com/hyperopt/hyperopt/blob/master/hyperopt/fmin.py#L156) of `fmin` is where the algorithm's `suggest` function is called and is given the `trials` object. This is the only way it takes into account previous runs, as `tpe.suggest` is just a function and has no internal state of its own. This function will return its suggested next place to look, which will then get appended to `trials` after being evaluated. This will keep looping until `max_evals` is reached.

[...]

PR #589, opened 2019-12-17
example of saving and loading trials

bjkomer:
Adding an example to the docs of how to save the trials object and continue the search from where it was left off. Addresses #267

### DIFF
--- a/docs/templates/getting-started/minimizing_functions.md
+++ b/docs/templates/getting-started/minimizing_functions.md
@@ -213,5 +213,5 @@ The basic technique involves:
 It's normal if this doesn't make a lot of sense to you after this short tutorial,
 but I wanted to give some mention of what's possible with the current code base,
 and provide some terms to grep for in the hyperopt source, the unit test,
-and example projects, such as [hyperopt-convnet](https://github.com/jaberg/hyperopt-convnet).
+and example projects, such as [hyperopt-convnet](https://github.com/hyperopt/hyperopt-convnet).
 Email me or file a github issue if you'd like some help getting up to speed with this part of the code.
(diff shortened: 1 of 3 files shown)


docs/templates/getting-started/minimizing_functions.md
```diff
@@ -181,4 +214,4 @@ It's normal if this doesn't make a lot of sense to you after this short tutorial
 but I wanted to give some mention of what's possible with the current code base,
 and provide some terms to grep for in the hyperopt source, the unit test,
 and example projects, such as [hyperopt-convnet](https://github.com/jaberg/hyperopt-convnet).
```
maxpumperla: I know this is not from your PR, but would you please replace all `jaberg` references by `hyperopt`? thanks!

maxpumperla:
@bjkomer thank you! this is much needed. just the little, unrelated nit and we're good to go here

docs/templates/getting-started/minimizing_functions.md
bjkomer: Good idea, they've been replaced now

maxpumperla:
@bjkomer awesome, thanks!

wxie2013:
where is the conclusion of all these discussion documented? The website: https://hyperopt.github.io/hyperopt/,  seems outdated.

wxie2013:
> where is the conclusion of all these discussion documented? The website: https://hyperopt.github.io/hyperopt/, seems outdated. By combing though gitbub, the latest doc seems to be the following which is dated 12/18/2019
> 
> https://github.com/bjkomer/hyperopt/blob/e017ddc60331631c7fc4a9208f1b3a2794849e88/docs/templates/getting-started/minimizing_functions.md
> 
> while in the official doc, it's dated Feb 2018.

also for the case the continuing the training from previous run, above threshold mention that one need to load the same rstate object while in the 12/18/2019 doc, it doesn't say so.  What should one do in this case? Maybe it doesn't matter to use either case because the starting point of the tries is the same?

PR #590, opened 2019-12-24
Remove unused param upper categorical

marctorrellas:
In preparation for adding a lower bound for randint, I'm doing this PR to simplify the categorical hyperparameter, where an `upper` parameter is not needed.

In addition to this, I refactored some of the code to hopefully make it more readable

### DIFF
--- a/hyperopt/tests/test_spark.py
+++ b/hyperopt/tests/test_spark.py
@@ -497,8 +497,8 @@ class FMinTestCase(unittest.TestCase, BaseSparkContext):
     def test_timeout_with_job_cancellation(self):
         if not self.sparkSupportsJobCancelling():
             print(
-                "Skipping timeout test since this Apache PySpark version does not support "
-                "cancelling jobs by job group ID."
+                "Skipping timeout test since this Apache PySpark version does not "
+                "support cancelling jobs by job group ID."
             )
             return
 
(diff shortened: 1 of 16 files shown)


hyperopt/tests/test_spark.py
```diff
@@ -488,8 +489,8 @@ def fn(x):
             self.assertIn(
                 "SparkTrials will block",
                 log_output,
-                """ "SparkTrials will block" missing from log: {log_output}""".format(
-                    log_output=log_output
+                "SparkTrials will block missing from log: log_output".format(
+                    log_output
```
maxpumperla: that change doesn't do what you intend, I think. It will just print literally "log_output". you need those `{}`.

maxpumperla:
btw, merry christmas @marctorrellas! you're very dedicated, I like it :D

hyperopt/tests/test_spark.py
marctorrellas: yup, I didn't remove them in purpose 😇 I reverted back to the original version with """, as the string generated had " inside of it and my intention was more about ensuring the max character length rather than changing anything

marctorrellas:
> btw, merry christmas @marctorrellas! you're very dedicated, I like it :D

haha merry xmas @maxpumperla 

I do what I can, when I have time. In January I'm starting a new job, so don't count on me by then, but I'll try to finish the work on randint if I manage to make it work

marctorrellas:
@maxpumperla, what do you think about assuming at least python3 (ideally 3.6 but I'd be happy at least with 3) for the next release?

maxpumperla:
@marctorrellas for the next minor release let's keep stuff as is. but I'm all for dropping python 2 support next year. let's do it!

PR #592, opened 2019-12-29
big refactoring + fix pchoice

marctorrellas:
I almost have randint with a lower bound working, but to get to it I refactored some stuff, found some things that can be done better (or don't work as expected), and also fixed pchoice. Ok, one thing at a time. This PR:

- Refactors some silly stuff, e.g. lines exceeding max char length, dicts declared using dict(), names for variables

- Adds some comments to hopefully understand better the code

- Refactors lot of code in tpe.py for the sake of understanding better what's going on. 

- Added some TODOs, especially important the one in tpe.py line 911. It looks like all points suggested are the same (?!), not sure this is correct

- I removed the old_bincount given that np is always > version 1.6, so no need for it. Also removed some other stuff

- I found out that tests for pchoice were failing something (depending on the random seed). I fixed one seed where it was always failing and found out that for some reason the p passed in pchoice is not [p(0), p(1) ... ] but something like  [ [p(0), p(1)...],  [p(0), p(1),...] , ...], i.e. a list of size len(observations) with lists of size Ncategories. So when doing len(p) for pchoice it passes the wrong parameter to bincount. I fixed it with a bit of a hack, but at least it does what it's supposed to do. Happy to hear alternatives

Once this is in, I will, hopefully soon, work on the tests for randint with a lower bound and push another PR :)

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -212,6 +212,8 @@ class FMinIter(object):
 
     def run(self, N, block_until_done=True):
         """
+        Run `self.algo` iteratively (use existing `self.trials` to produce the new
+        ones), update, and repeat
         block_until_done  means that the process blocks until ALL jobs in
         trials are not in running or new state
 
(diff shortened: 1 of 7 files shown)


marctorrellas:
finally, ready to review @maxpumperla

hyperopt/fmin.py
```diff
@@ -271,17 +278,11 @@ def get_n_unfinished():
...
-                            d["result"]["loss"]
-                            for d in self.trials.trials
-                            if d["result"]["status"] == "ok"
-                        ]
-                    )
-                    progress_ctx.postfix = "best loss: " + str(best_loss)
-                except:
-                    pass
+
+                # update progress bar with the min loss among trials with status ok
```
maxpumperla: nice! :+1:

maxpumperla:
@marctorrellas thanks. unfortunately I don't have any insights on your TODOs there. @jaberg will have to comment on these

PR #593, opened 2019-12-31
fmin parameter: loss_threshold, can work independently, or with max_evals

weixingsun:
I tested it with my simple test cases, please help to review it
And I am not quite sure what to do if someone change the loss function

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -14,6 +14,7 @@ from timeit import default_timer as timer
 import numpy as np
 
 from hyperopt.base import validate_timeout
+from hyperopt.base import validate_confidence
 from . import pyll
 from .utils import coarse_utcnow
 from . import base
@@ -118,6 +119,7 @@ class FMinIter(object):
         poll_interval_secs=1.0,
         max_evals=sys.maxsize,
         timeout=None,
+        confidence=None,
         verbose=False,
         show_progressbar=True,
     ):
@@ -138,6 +140,7 @@ class FMinIter(object):
         self.max_queue_len = max_queue_len
         self.max_evals = max_evals
         self.timeout = timeout
+        self.confidence = confidence
         self.start_time = timer()
         self.rstate = rstate
         self.verbose = verbose
@@ -237,9 +240,11 @@ class FMinIter(object):
         ) as progress_ctx:
 
             all_trials_complete = False
-            while (n_queued < N or (block_until_done and not all_trials_complete)) and (
-                self.timeout is None or (timer() - self.start_time) < self.timeout
-            ):
+            best_loss = -1
+            quit_loss = 100-self.confidence
+            while (n_queued < N or (block_until_done and not all_trials_complete)) 
+              and ( self.timeout is None or (timer() - self.start_time) < self.timeout )
+              and (self.confidence is None or best_loss > quit_loss):
                 qlen = get_queue_len()
                 while (
                     qlen < self.max_queue_len and n_queued < N and not self.is_cancelled
@@ -328,6 +333,7 @@ def fmin(
     algo,
     max_evals,
     timeout=None,
+    confidence=None,
     trials=None,
     rstate=None,
     allow_trials_fmin=True,
@@ -380,6 +386,10 @@ def fmin(
     timeout : None or int, default None
         Limits search time by parametrized number of seconds.
         If None, then the search process has no time constraint.
+
+    confidence : None or double, default None
+        Limits search time when minimal loss reduced to certain amount.
+        If None, then the search process has no best_loss constraint.
 
     trials : None or base.Trials (or subclass)
         Storage for completed, ongoing, and scheduled evaluation points.  If
@@ -447,6 +457,7 @@ def fmin(
             rstate = np.random.RandomState()
 
     validate_timeout(timeout)
+    validate_confidence(confidence)
 
     if allow_trials_fmin and hasattr(trials, "fmin"):
         return trials.fmin(
@@ -455,6 +466,7 @@ def fmin(
             algo=algo,
             max_evals=max_evals,
             timeout=timeout,
+            confidence=confidence,
             max_queue_len=max_queue_len,
             rstate=rstate,
             pass_expr_memo_ctrl=pass_expr_memo_ctrl,
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -242,7 +242,7 @@ class FMinIter(object):
             all_trials_complete = False
             best_loss = -1
             quit_loss = 100-self.confidence
-            while (n_queued < N or (block_until_done and not all_trials_complete)) 
+            while (n_queued < N or (block_until_done and not all_trials_complete))
               and ( self.timeout is None or (timer() - self.start_time) < self.timeout )
               and (self.confidence is None or best_loss > quit_loss):
                 qlen = get_queue_len()
@@ -491,6 +491,7 @@ def fmin(
         trials,
         max_evals=max_evals,
         timeout=timeout,
+        confidence=confidence,
         rstate=rstate,
         verbose=verbose,
         max_queue_len=max_queue_len,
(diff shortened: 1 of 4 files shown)


weixingsun:
When set max_evals=sys.maxsize, without confidence defined, then it would be a infinite loop, does it break some other test cases, and cause the merge failure?

marctorrellas:
you need to check the tests that are failing. If you cannot run the tests locally, go to "Details" above, get into travis, click on the distro failing (in this case all python versions are failing) and check the "job log" for the error. In this case it's a SyntaxError, so should be easy to fix

In general, if you can run the tests locally, less spam you create here. You can do so as e.g.

`python -m nose -s hyperopt/tests/test_ipy.py`

for example, so that you just run one test and the "-s" shows all prints. Also, if you write breakpoints with pdb you can debug. To run all tests you can do e.g.

`bash run_tests.sh hyperopt/tests/`

In some cases (not in this case), some tests depend on the randomness of the algorithms and fail irresp. of you doing something wrong

hyperopt/fmin.py
```diff
@@ -381,6 +387,10 @@ def fmin(
         Limits search time by parametrized number of seconds.
         If None, then the search process has no time constraint.
 
+    confidence : None or double, default None
+        Limits search time when minimal loss reduced to certain amount.
+        If None, then the search process has no best_loss constraint.
```
marctorrellas: I think the term confidence is used for other concepts in ML/maths, so I would go with "tolerance", "loss_threshold", "loss_cutoff", or similar. About the docstring, what do you think about

```Limits search when loss reaches at least the max_loss/tolerance/whatever value.
        If None, then the search process has no constraint on the loss value.```

Comment (weixingsun) [reply hyperopt/fmin.py]: I would prefer loss_threshold, will change it soon

**Comment by weixingsun** [comment]
@marctorrellas happy new year, I don't understand why my test passed, but CI check failed with TypeError: fmin() got an unexpected keyword argument 'loss_threshold'

**Comment by marctorrellas** [comment]
Your contribution is very similar to the one for timeout, so please try to look at the changes in that other PR to fix your tests. If you don't manage to fix it I will take a look

**Comment by weixingsun** [comment]
@marctorrellas thanks for your reply, looking at build #211, I don't know why 2.7 and 3.6 passed, but 3.5 failed.

**Comment by marctorrellas** [comment]
yeah, these are flaky tests. Run the tests again by closing and reopening the PR. Ideally, we would fix the tests though...

**Comment by marctorrellas** [comment]
do you mind changing the PR title to e.g.  "fmin parameter: loss threshold ..." or similar?

File: hyperopt/fmin.py [hunk 2]
```diff
@@ -14,6 +14,7 @@
 import numpy as np
 
 from hyperopt.base import validate_timeout
+from hyperopt.base import validate_loss_threshold
```
marctorrellas: please merge with previous line

hyperopt/fmin.py
```diff
@@ -237,9 +240,10 @@ def get_n_unfinished():
         ) as progress_ctx:
 
             all_trials_complete = False
-            while (n_queued < N or (block_until_done and not all_trials_complete)) and (
-                self.timeout is None or (timer() - self.start_time) < self.timeout
-            ):
+            best_loss = sys.maxsize
+            while ( ( n_queued < N or (block_until_done and not all_trials_complete) )
+              and   ( self.timeout is None or (timer() - self.start_time) < self.timeout )
+              and   ( self.loss_threshold is None or best_loss > self.loss_threshold) ):
```
marctorrellas: I think this set of conditions is getting tricky to read, can you add a comment above it to explain (not only your part but all of it)?

I would use >= rather than > for users who want to minimise something to exactly a number, e.g. 0

minor suggestion: use float("inf")  for initial value of best_loss?

hyperopt/fmin.py
marctorrellas: in the first sentence you talk about "minimal loss" while in the first one about the "best_loss". I think it's a bit confusing. What about something like:

```Limits search time when the loss for the best trial is lower or equal than this value.
        If None, then the search process has no constraint on the loss, and will stop
        based on other parameters, e.g. `max_evals`, `timeout`.```

File: hyperopt/spark.py [hunk 4]
```diff
@@ -213,6 +217,11 @@ def fmin(
                 validate_timeout(timeout)
                 self.timeout = timeout
 
+        if loss_threshold is not None:
+            if self.loss_threshold is None:
+                validate_loss_threshold(loss_threshold)
+                self.loss_threshold=loss_threshold
```
marctorrellas: why do you need these two if's separated here?

hyperopt/tests/test_fmin.py
```diff
@@ -291,3 +291,42 @@ def test_invalid_timeout():
...
+    loss_threshold=0.001
+    hypopt_trials = Trials()
+    best = fmin(
+        fn=lambda x: x ** 2,
+        space=hp.uniform('x', -10, 10),
+        loss_threshold=loss_threshold,
+        algo=rand.suggest,
+        trials=hypopt_trials,
+        return_argmin=False,
+        #max_evals=100
```
marctorrellas: please set a rstate so tests are (more) deterministic, and remove the commented line if not needed

hyperopt/tests/test_fmin.py
```diff
@@ -291,3 +291,42 @@ def test_invalid_timeout():
...
+
+
+def test_invalid_loss_threshold():
+    fn = lambda x: [time.sleep(1), x][1]
+    space = hp.choice("x", range(20))
+
+    for wrong_loss_threshold in ["", True]:
+        expected_message = "The loss_threshold argument should be None or a numeric value. Given value: {m}".format(
+            m=wrong_loss_threshold
+        )
```
marctorrellas: please use black as recommended in the dev guidelines so that the max char length is no more than 88 chars

hyperopt/fmin.py
```diff
@@ -237,9 +239,12 @@ def get_n_unfinished():
...
             all_trials_complete = False
-            while (n_queued < N or (block_until_done and not all_trials_complete)) and (
-                self.timeout is None or (timer() - self.start_time) < self.timeout
-            ):
+            best_loss = float("inf")
+            while ( 
+                    ( n_queued < N or (block_until_done and not all_trials_complete) )     # more run to Q     || ( block_flag & trials not done )
+              and   ( self.timeout is None or (timer() - self.start_time) < self.timeout ) # no timeout        || < current last time
+              and   ( self.loss_threshold is None or best_loss >= self.loss_threshold)     # no loss_threshold || < current best_loss
+                  ):
```
marctorrellas: please rewrite this so that you're not exceeding the 88 max char line length

hyperopt/spark.py
```diff
@@ -49,7 +49,7 @@ class SparkTrials(Trials):
     # Hard cap on the number of concurrent hyperopt tasks (Spark jobs) to run. Set at 128.
     MAX_CONCURRENT_JOBS_ALLOWED = 128
 
-    def __init__(self, parallelism=None, timeout=None, spark_session=None):
+    def __init__(self, parallelism=None, timeout=None, loss_threshold=None, spark_session=None):
```
marctorrellas: as recommended in the dev guidelines, please use black to ensure you don't exceed the 88 chars

hyperopt/tests/test_fmin.py
```diff
@@ -291,3 +291,42 @@ def test_invalid_timeout():
...
+
+
+def test_invalid_loss_threshold():
+    fn = lambda x: [time.sleep(1), x][1]
+    space = hp.choice("x", range(20))
+
+    for wrong_loss_threshold in ["abc", True]:
+        expected_message = "The loss_threshold argument should be None or a numeric value. Given value: {m}".format(
+            m=wrong_loss_threshold
+        )
```
marctorrellas: please adjust to 88 chars

maxpumperla:
@WeichenXu123 @marctorrellas are we good to go here?

marctorrellas:
> @WeichenXu123 @marctorrellas are we good to go here?

👍

marctorrellas:
reminder to merge @maxpumperla

maxpumperla:
@marctorrellas cheers! ;)

PR #595, opened 2020-01-06
Add lower bound randint

marctorrellas:
After the two PRs to prepare this, finally we can add a lower bound to randint

Given that randint was using categorical, I added a lower bound by considering a shifted categorical. While there are some hacks and it could be way nicer, it seems to work fine (see added tests)

Comments about the approach or anything, welcome :)

### DIFF
--- a/hyperopt/atpe.py
+++ b/hyperopt/atpe.py
@@ -1,5 +1,8 @@
 """
-    Implements the ATPE algorithm. See https://www.electricbrain.io/blog/learning-to-optimize and https://www.electricbrain.io/blog/optimizing-optimization to learn more
+    Implements the ATPE algorithm. See
+    https://www.electricbrain.io/blog/learning-to-optimize
+    and
+    https://www.electricbrain.io/blog/optimizing-optimization to learn more
 """
 
 __authors__ = "Bradley Arsenault"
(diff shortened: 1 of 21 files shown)


maxpumperla:
@marctorrellas I can't get the python2 build to run? does that work locally for you (I know, I know, we'll throw it out eventually, hang tight)

marctorrellas:
> @marctorrellas I can't get the python2 build to run? does that work locally for you (I know, I know, we'll throw it out eventually, hang tight)

it was working locally for me 2 or 3 commits before this. Apart from the new tests, I don't see a reason for it not to work. Let's restart travis and see what errors we found. What errors you observe?

marctorrellas:
working :) there was a problem with the previous PR about loss_threshold...

hyperopt/atpe.py
```diff
@@ -488,8 +489,10 @@ def createHyperparameterConfigForHyperoptDomain(domain):
             return data
         elif domain.name == "randint":
             data = {"type": "number"}
-            max = domain.pos_args[0]._obj
-            data["max"] = max
+            low = domain.pos_args[0]._obj
+            high = domain.pos_args[1]._obj
+            data["min"] = 0 if high is None else low
+            data["max"] = high or low
```
jawadMansoor, line 495: shouldn't this be
`data["min"] =  0 if low is None else low`
`data["max"] = high`

and in this case high will be required parameter and low will be optional, currently in this code, if I am not wrong, low is treated as mandatory argument and it means upper bound (high) in case low is not given... and it makes less sense.

PR #600, opened 2020-01-11
[ML-9328] Support dynamic allocation in hyperopt integration

WeichenXu123:
Support dynamic allocation in hyperopt integration. Only throw a warning when current task slots is smaller than requested parallelism.

When user specify parallelism to be greater than `max_num_concurrent_tasks`, print a warning notify user that this will trigger spark executor dynamic allocation (if spark configured with dynamic allocation mode).

When user do not specify parallelism, it will be set to `max(spark_default_parallelism, max_num_concurrent_tasks)` as default behavior, and print a deprecation warning, next released version will remove this default behavior.

### DIFF
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -4,6 +4,7 @@ import copy
 import threading
 import time
 import timeit
+import warnings
 
 from hyperopt import base, fmin, Trials
 from hyperopt.base import validate_timeout, validate_loss_threshold
@@ -91,8 +92,11 @@ class SparkTrials(Trials):
         )
         # maxNumConcurrentTasks() is a package private API
         max_num_concurrent_tasks = self._spark_context._jsc.sc().maxNumConcurrentTasks()
+        spark_default_parallelism = self._spark_context.defaultParallelism
         self.parallelism = self._decide_parallelism(
-            max_num_concurrent_tasks, parallelism
+            requested_parallelism=parallelism,
+            spark_default_parallelism=spark_default_parallelism,
+            max_num_concurrent_tasks=max_num_concurrent_tasks
         )
 
         if not self._spark_supports_job_cancelling and timeout is not None:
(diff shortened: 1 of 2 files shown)


maxpumperla:
@WeichenXu123 mind explaining this a bit more? for now I only see a few logging changes? in any case, let me know once you're good to go. thanks

hyperopt/spark.py
```diff
@@ -126,20 +126,19 @@ def _decide_parallelism(max_num_concurrent_tasks, parallelism):
...
-                "User-specified parallelism ({p}) is greater than "
-                "max_num_concurrent_tasks ({c}), so parallelism will be set to "
-                "max_num_concurrent_tasks.".format(
+                "User-specified parallelism ({p}) is greater than the max number of concurrent "
+                "tasks ({c}) this cluster can run now. If dynamic allocation is enabled for the "
+                "cluster, you might see more executors allocated.".format(
                     p=parallelism, c=max_num_concurrent_tasks
                 )
             )
-            parallelism = max_num_concurrent_tasks
```
WeichenXu123: @maxpumperla Not only logging changed. See this line, I remove the parallelism limit here. So that, when spark dynamic allocation enabled, when user-specified parallelism larger than current spark max number of concurrent tasks, it will allow spark to allocate more executors (The previous code will limit parallelism to be max_num_concurrent_tasks, which prevent spark to allocate more executors)

hyperopt/spark.py
```diff
@@ -126,20 +126,19 @@ def _decide_parallelism(max_num_concurrent_tasks, parallelism):
         elif parallelism <= 0:
             logger.warning(
                 "User-specified parallelism was invalid value ({p}), so parallelism "
```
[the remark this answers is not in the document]
WeichenXu123: I use default parallelism = `max(sc.defaultParallelism, max_num_concurrent_tasks)` and print deprecation message when using default parallelism.

WeichenXu123:
@mengxr @maxpumperla Any more comments ? Thanks!

mengxr:
I will push some changes directly to your branch.

mengxr:
@maxpumperla This PR should be ready for merge. PTAL?

mengxr:
Ping @maxpumperla :)

hyperopt/spark.py
```diff
@@ -111,43 +116,46 @@ def __init__(self,
         self.refresh()
 
     @staticmethod
-    def _decide_parallelism(max_num_concurrent_tasks, parallelism):
+    def _decide_parallelism(
+            requested_parallelism,
+            spark_default_parallelism,
+            max_num_concurrent_tasks):
```
marctorrellas: please use black as the dev guidelines suggest :)

mengxr:
@marctorrellas I updated the source code with black. Now real code changes are mixed with code style changes. Why not running black on all source and then put the linter in CI? cc: @maxpumperla

marctorrellas:
that cannot be done until we stop supporting python < 3.6. Our current CI has 3 builds: 2.7, 3.5 and 3.6

mengxr:
@marctorrellas @maxpumperla Could you take another look?

hyperopt/spark.py
```diff
@@ -111,43 +113,50 @@ def __init__(self,
...
-            parallelism = max_num_concurrent_tasks
-        elif parallelism <= 0:
             logger.warning(
-                "User-specified parallelism was invalid value ({p}), so parallelism "
-                "will be set to max_num_concurrent_tasks ({c}).".format(
-                    p=parallelism, c=max_num_concurrent_tasks
-                )
+                "The cluster has no executors currently. "
+                "The trials won't start until some new executors register."
             )
```
marctorrellas: should this just show a warning or raise a ValueError?

hyperopt/tests/test_spark.py
```diff
@@ -273,91 +273,86 @@ def test_parallelism_arg(self):
...
-
-        # Given valid parallelism
-        parallelism = SparkTrials._decide_parallelism(max_num_concurrent_tasks, None)
-        self.assertEqual(
-            parallelism,
-            max_num_concurrent_tasks,
-            "The default parallelism ({p}) did not equal max_num_concurrent_tasks"
-            " ({e})".format(p=parallelism, e=max_num_concurrent_tasks),
-        )
+            # Test requested_parallelism is None or negative values.
```
marctorrellas: maybe a different test (same file but a different function) would help to localise where the tests are failing. If everything is in the same is not that easy. If we had pytest this would be a bit nicer using pytest.mark.parametrize....🤷‍♂️

hyperopt/tests/test_spark.py
```diff
@@ -273,91 +273,86 @@ def test_parallelism_arg(self):
...
-            self.assertIn(
-                "SparkTrials.MAX_CONCURRENT_JOBS_ALLOWED ({c})".format(
-                    c=SparkTrials.MAX_CONCURRENT_JOBS_ALLOWED
-                ),
-                log_output,
-                """MAX_CONCURRENT_JOBS_ALLOWED value missing from log: {log_output}""".format(
-                    log_output=log_output
-                ),
-            )
+            # Test requested_parallelism exceeds hard cap
```
marctorrellas: and another separate test here as well

maxpumperla:
I'm OK with this as is. thanks for this contribution

Issue #637, opened 2020-03-13
building from souces install docs instead of hyperopt


smuzaffar:
We are trying to install hyperopt==0.2.2 from souces but looks like it install docs instead , see the log below
```
+ pip3 install --upgrade --upgrade-strategy only-if-needed --no-cache-dir --disable-pip-version-check --user -v --no-clean --no-deps hyperopt-0.2.2.tar.gz
Processing ./hyperopt-0.2.2.tar.gz
  Running setup.py (path:/build/muz/p3/w/BUILD/slc7_amd64_gcc820/external/py2-hyperopt/0.2.2-cms/cmsdist-tmp/pip-ynd_ae9g-build/setup.py) egg_info for package from file:///build/muz/p3/w/BU
ILD/slc7_amd64_gcc820/external/py2-hyperopt/0.2.2-cms/hyperopt-0.2.2.tar.gz
    Running command python setup.py egg_info
    DEBUG:root:skipping submodule hyperopt/pyll
    DEBUG:root:skipping submodule hyperopt/tests
    DEBUG:root:skipping submodule hyperopt.pyll/tests
    running egg_info
    creating pip-egg-info/docs.egg-info
    writing pip-egg-info/docs.egg-info/PKG-INFO
    writing dependency_links to pip-egg-info/docs.egg-info/dependency_links.txt
    writing entry points to pip-egg-info/docs.egg-info/entry_points.txt
    writing requirements to pip-egg-info/docs.egg-info/requires.txt
    writing top-level names to pip-egg-info/docs.egg-info/top_level.txt
    writing manifest file 'pip-egg-info/docs.egg-info/SOURCES.txt'
    reading manifest file 'pip-egg-info/docs.egg-info/SOURCES.txt'
    writing manifest file 'pip-egg-info/docs.egg-info/SOURCES.txt'
  Source in ./cmsdist-tmp/pip-ynd_ae9g-build has version 0.2.2, which satisfies requirement docs==0.2.2 from file:///build/muz/p3/w/BUILD/slc7_amd64_gcc820/external/py2-hyperopt/0.2.2-cms/h
yperopt-0.2.2.tar.gz
Installing collected packages: docs
....
...
    creating build/lib/hyperopt/pyll/tests
    copying hyperopt/pyll/tests/test_stochastic.py -> build/lib/hyperopt/pyll/tests
    copying hyperopt/pyll/tests/__init__.py -> build/lib/hyperopt/pyll/tests
[...]

Issue #613, opened 2020-02-04
Inconsistent package name from setup.py


dc-stripe:
Occasionally when our CI builds the whl for `hyperopt-0.2.2`, the resulting file is called `docs` instead of `hyperopt`. I believe, but am not certain, that what's happening is that `setuptools.find_packages()` sometimes returns ["docs", "hyperopt"] and so this code in `setup.py` provides `docs`:
```
if package_name is None:
    package_name = packages[0]
```

PR #638, opened 2020-03-17
Setup simplification & fixes

akx:
* Remove distribute_setup: a regular setuptools setup is all that's necessary these days
* Remove custom resource-finding code from setup.py
  * Instead use MANIFEST.in as usual
  * This also fixes the problem that the package could arbitrarily be called `docs` (#613, #637)
* Remove redundant hyperopt-mongo-worker script (installed by entrypoints instead)
* Mark the built wheel as py2.py3


maxpumperla:
@akx makes perfect sense, thank you!

PR #644, opened 2020-03-30
Remove test of method which no longer exists.

magdapoppins:
Remove this line since `_bincount_slow()` has been removed and will cause tests to fail.


maxpumperla:
@magdapoppins keep 'em coming :D

PR #648, opened 2020-04-08
Some Refactoring

marload:
First of all, thank you so much for creating such a wonderful project!

I found a code that required some refactoring while using hyperopt.
So I modified the codes.

Thank you again for your efforts!
It's a really, really great project.


maxpumperla:
looking good @marload, thanks for the simplifications

marload:
It's a really an awesome project! Thx

PR #652, opened 2020-04-14
[ML-10384] Replace the retry warnings for long running tasks in SparkTrials

liangz1:
### Problem
In SparkTrials, the long-running tasks will trigger the following warning message:
```
Consider setting `spark.conf.set('spark.task.maxFailures', '1')` to prevent retries.
```
However, this spark conf cannot be changed after the spark session is initialized, and it's tricky to find the correct way to set it. If the user simply follows the instruction in the warning messages and update the spark conf in their code, they would get:
```
org.apache.spark.sql.AnalysisException: Cannot modify the value of a Spark config: spark.task.maxFailures;
```
### Proposed fix
1. Remove the `Consider setting spark.conf.set('spark.task.maxFailures', '1')` warning.
2. Capture the exception in `run_task_on_executor()`, handle the exception and return normally. This is to prevent Spark from retrying the failed task.
3. Refactor the code for `finish_trial_run` logic.

### Test
unittest

### DIFF
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -292,9 +292,6 @@ class _SparkFMinState:
     Each trial's thread runs 1 Spark job with 1 task.
     """
 
-    # definition of a long-running trial, configurable here for testing purposes
-    _LONG_TRIAL_DEFINITION_SECONDS = 60
-
     def __init__(self, spark, eval_function, space, trials):
         self.spark = spark
         self.eval_function = eval_function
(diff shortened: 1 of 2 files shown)


WeichenXu123:
Looks good. Let's add unit test.

mengxr:
@liangz1 @WeichenXu123 I don't think it is reliable to use running time to tell whether it is a real Spark exception or a user exception. Shall we make the logic simpler by raising SparkException for any failures? Should verify that hyperopt would treat failures gracefully.

WeichenXu123:
> @liangz1 @WeichenXu123 I don't think it is reliable to use running time to tell whether it is a real Spark exception or a user exception. Shall we make the logic simpler by raising SparkException for any failures? Should verify that hyperopt would treat failures gracefully.

Looks reasonable. We can simplify logic by: In runtime, if it do not timeout and should retry, we can raise SparkException, otherwise we can swallow the exception and return the exception in the task.

liangz1:
@maxpumperla This PR is ready for review. Can you take a look? Thanks!

maxpumperla:
@liangz1 @liangz1 @mengxr just created a new release (0.2.4) for you as well

Issue #449, opened 2018-12-17
Feature request: add stopping condition to fmin


faye36:
It would be useful to add a way to stop the hyperparameter search early after evaluating one or more trials  where the loss has fallen below a certain threshold - instead of always continuing the search until max_evals is reached.

ninnghazad:
the way i do that is to loop fmin with max_evals=len(trials)+batch_size and check conditions after each batch. batch_size i usually choose by rule-of-thumb * number-of-workers or something like that.

PhilipMay:
What do you think about a timed stopping condition? Instead of telling hyperopt to run for a certain number of evaluations you could tell hyperapt to run x minutes?

ninnghazad:
that'd be a nice addition - can't cleanly do that from the outside. however, the loss-based option also could have the advantage over external solutions that (i think) the state of the tpe would not be interrupted to check on the stopping condition.

PhilipMay:
What about a time based early stopping? If loss does not improve after x minutes then leave it. Or a mix of time and iterations...

maxpumperla:
I think the loss-based stopping is more useful in practice, but why not give users options? We could think about a simple `stopping_criterion` argument to fmin that is None by default and takes a function that evaluates as boolean to tell fmin when to stop. We can provide a few default implementations, like time- or loss-based.

PhilipMay:
> I think the loss-based stopping is more useful in practice, but why not give users options? We could think about a simple `stopping_criterion` argument to fmin that is None by default and takes a function that evaluates as boolean to tell fmin when to stop. We can provide a few default implementations, like time- or loss-based.

Great suggestion. Will have a look into it.

maxpumperla:
@PhilipMay just a rough idea, may not even work as suggested. but if you come up with a good idea, I'm happy to discuss, as you know.

jasonminsookim:
> the way i do that is to loop fmin with max_evals=len(trials)+batch_size and check conditions after each batch. batch_size i usually choose by rule-of-thumb * number-of-workers or something like that.

Could you provide example code for this?

ninnghazad:
@jasonminsookim something like this. it processes trials in chunks, args.trials can be a filename (serial execution) or a mongo-url (parallel execution):
```
                if args.trials.startswith("mongo://"):
			trials = MongoTrials(args.trials, exp_key=args.exp_key)
			mongomode = True
		else:
			mongomode = False
			try:
[...]

PR #661, opened 2020-04-30
[#449] add early stop parameter and default implementation

QuentinAmbard:
this PR add support for an early_stop parameter in the fmin function.
Hyperopt will stop the run when this function returns True.

Default implementation allow to stop the search if the loss doesn't increase after X run by a percent of X.
Custom implementations can also be added as stop function

### DIFF
--- a/hyperopt/early_stop.py
+++ b/hyperopt/early_stop.py
@@ -1,0 +1,35 @@
+import logging
+import math
+
+logger = logging.getLogger(__name__)
+
+
+def no_progress_loss(iteration_stop_count=20, percent_increase=0.0):
+    """
+    Stop function that will stop after X iteration if the loss doesn't increase
+
+    Parameters
+    ----------
+    iteration_stop_count: int
+        search will stop if the loss doesn't improve after this number of iteration
+    percent_increase: float
+        allow this percentage of variation within iteration_stop_count.
+        Early stop will be triggered if the data didn't change for more than this number
+        during iteration_stop_count rounds
+    """
+    def stop_fn(trials, best_loss=None, iteration_no_progress=0):
+        new_loss = trials.trials[len(trials.trials)-1]['result']['loss']
+        if best_loss == None:
+            return False, [new_loss, iteration_no_progress+1]
+        best_loss_threshold = best_loss - abs(best_loss * (percent_increase/100.0))
+        if new_loss < best_loss_threshold:
+            best_loss = new_loss
+            iteration_no_progress = 0
+        else:
+            iteration_no_progress += 1
+            logger.debug("No progress made: %d iteration on %d. best_loss=%.2f, best_loss_threshold=%.2f, new_loss=%.2f"
+                         % (iteration_no_progress, iteration_stop_count, best_loss, best_loss_threshold, new_loss))
+
+        return iteration_no_progress >= iteration_stop_count, [best_loss, iteration_no_progress]
+
+    return stop_fn
--- a/hyperopt/early_stop.py
+++ b/hyperopt/early_stop.py
@@ -1,5 +1,4 @@
 import logging
-import math
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


maxpumperla:
@QuentinAmbard looks pretty good! thanks. some of the tests seem to fail due to the new keyword, can you confirm this locally?

```
======================================================================
ERROR: test_accepting_sparksession (hyperopt.tests.test_spark.FMinTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/hyperopt/hyperopt/hyperopt/tests/test_spark.py", line 260, in test_accepting_sparksession
    trials=spark_trials,
  File "/home/travis/build/hyperopt/hyperopt/hyperopt/fmin.py", line 503, in fmin
    early_stop=early_stop,
TypeError: fmin() got an unexpected keyword argument 'early_stop'
```

QuentinAmbard:
I fixed the error, all the tests are passing on my laptop (python 3.7). However there is still an error with python 2.5 and 3.5 in Travis:

`...OSError: [Errno 17] File exists: '/home/travis/build/hyperopt/hyperopt/mongoexp_test_dir'`
and
`...FileNotFoundError: [Errno 2] No such file or directory: 'home.inuse'`

I don't see what I could have break based on the error stack. @maxpumperla do you have an idea ? Thanks

maxpumperla:
@QuentinAmbard that seems unrelated to your changes... not sure what's happening. Strangely, I can't restart the build on travis.org right now.

maxpumperla:
@QuentinAmbard I guess the typo must have caused the issue ;)

Thank you very much for this PR!

ciberger:
Hi @QuentinAmbard, thanks a lot for your PR! Could you provide an example of how to use this new feature?

QuentinAmbard:
Sure, you can just use it like this:

 ```
  from hyperopt.early_stop import no_progress_loss
   trials = generate_trials_to_calculate([{'x': -100}])
    fmin(
        fn = lambda x: x,
        space=hp.uniform("x", -5, 5),
        algo=rand.suggest,
        max_evals=500,
        trials=trials,
        early_stop_fn=no_progress_loss(iteration_stop_count=20, percent_increase=0.0)
    )
```

**Comment by Pked01** [comment]
@QuentinAmbard  loss should decrease accuracy should increase, 
shouldn't parameter percent_increase should be replaced with percent_decrease??

**Comment by QuentinAmbard** [comment]
Yes that's correct, something like `percent_increase` or `percent_change` is probably a better naming...

@maxpumperla what do you think, should I create a small PR to change the variable name in the function?

fyi It's computed like this:

```
        best_loss_threshold = best_loss - abs(best_loss * percent_increase/100.0)
        if new_loss < best_loss_threshold:
                 #progress made
        else:
                 #no progress
```

File: hyperopt/early_stop.py [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+    ----------
+    iteration_stop_count: int
+        search will stop if the loss doesn't improve after this number of iteration
+    percent_increase: float
+        allow this percentage of variation within iteration_stop_count.
+        Early stop will be triggered if the data didn't change for more than this number
+        after iteration_stop_count rounds
+    """
+    def stop_fn(trials, best_loss=None, iteration_no_progress=0):
+        new_loss = trials.trials[len(trials.trials)-1]['result']['loss']
```
vperiyasamy, line 20: Does this criteria fail when spark trials are used in parallel? I'm getting `KeyError: 'loss'` when trying this out, I believe because this callback was executed after one trial finished, but `trials.trials[len(trials.trials)-1]` doesn't refer to that trial

jawadMansoor:
I have faced the same issue and I was about to make a separate issue. When using sparkTrials in fmin() and early stopping criteria we get keyError: 'loss'. No error is raised if we use base.trails()...

Issue #662, opened 2020-05-02
Poor treatment of losses with value nan


GiliR4t1qbit:
Sometimes your loss might end up being nan. It turns out that hyperopt does not handle this well currently. I discovered this since for my algorithm, the first loss is always nan, and the rest are normally not. I noticed conflicting messages, first "TPE using 1/2 trials with best loss 0.000478" but the in the progress bar "best loss: nan". The latter is wrong, the former is right. I've figured out what the issue is - it's a subtle issue related to the way `min()` and `numpy.nan` interact (spoiler alert: never use `min()` if you might encounter a `numpy.nan`). The source of the issue is that the progress bar uses `min(losses)`, which is not safe when `losses` includes `numpy.nan` (see a detailed explanation below).

**Minimal Reproducer**
Here's a minimal reproducer. It's an objective that returns `NaN` as the first loss, and 1 after that:
```
import numpy

def objective(args):
    objective.n += 1
    
    if objective.n == 1:
        return numpy.nan
    else:
        return 1

from hyperopt import hp
space = hp.lognormal('unused', 0, 1)

from hyperopt import fmin, tpe

objective.n = 0
best = fmin(objective, space, algo=tpe.suggest, max_evals=100)

print(best)
```
The output I get is:
```100%|██████████| 100/100 [00:00<00:00, 591.91it/s, best loss: nan]```

If we instead switch the `nan` to occur in the second place by replacing `if objective.n == 1:` with `if objective.n == 2:`, then we get:
```
100%|██████████| 100/100 [00:00<00:00, 388.64it/s, best loss: 1.0]
```
So, the best loss reported is correct. The fact that it matters where the nan occurs is explained below. 

**Why We Should Avoid Using min() with nan**
Let's check what happens when at least one loss has the value `numpy.nan`. Since the expression `min(losses)` occurs in two places in the code, and the expression `np.argmin(losses)` occurs in one place, let's check directly how they react to a `numpy.nan`.

The first thing to note is the very surprising fact that for `min()` it matters _where the nan is_... Let's first try putting a nan value in the first place: 
```
l = [numpy.nan, 0, 1, 2]
min(l)
```
What do you think the result will be? I would have thought it would be `0,` but instead we get `nan`. 

Now let's try moving the nan one up:
```
l = [1, numpy.nan, 0, 2]
min(l)
```
This time we get `0`, the result I would have expected. So, what's happening? **Surely the order of the elements should not matter when using `min()`?!** 

[...]

PR #663, opened 2020-05-02
Deal with nan loss by replacing min->nanmin, argmin->nanargmin

GiliR4t1qbit:
Closes issue #662: 
- Currently the best loss can be reported wrongly in the progress bar due to relying on min() which is not safe when nans can occur (see issue for detailed description)
- Deal with the above by replacing `min` -> `nanmin`, `argmin` -> `nanargmin`, `if loss is not None` -> `if loss is not None and not np.isnan(loss)`.


maxpumperla:
@GiliR4t1qbit thank you!

richachoudhary:
@GiliR4t1qbit 
This helped.

PR #674, opened 2020-06-04
Code Formatting & Early Return Pattern Change

marload:
Early Return Pattern recommends 'if if return' more than 'if elif else'.

Thanks You!

This commit passed the pytest.


maxpumperla:
thanks! congrats to your first PR around here :D

PR #683, opened 2020-06-05
Change to Python3 Compatible

marload:
The notebook writed with existing python2 has been changed to python3 compatible. Thank you.


marload:
@maxpumperla 
I think TravisCI is broken. Can you check it?

PR #687, opened 2020-06-14
docs: fixes Spark parallelism error

gulyasm:
# What is this PR is about?
There is an error in the Spark scaleout documentation about what `parallelism` defaults to:
What it says: Defaults to the number of Executors.
What it should say: It doesn't default to the number of executors, but to `SparkContext.defaultParallelism`, which is the number of slots. 

# How does this PR solves the problem?
It fixes the documentation to contain the correct description. 

# Risks of this PR
**Negligable**
Documentation helps users to understand what is happening. Wrong documentation mislead users, hard to understand documentation discourage users from using the library. 
This is a very small change, risks are very limited.


maxpumperla:
@gulyasm thanks so much!

Issue #657, opened 2020-04-27
hyperopt.__version__ needs update


liangz1:
https://github.com/hyperopt/hyperopt/blob/1d438e5e96a7cc2a839dd2ca3aeaffc64e144976/hyperopt/__init__.py#L43

marctorsoc:
do you mind doing a quick PR for this @liangz1 ? is there a way to automate this @maxpumperla  to be updated every time we release?

maxpumperla:
@marctorsoc I'm honestly not sure where this is needed anymore. I suspect that `setup.py` was picking up this info at some point. Personally I've never used dunder version this way.

marctorsoc:
it's really weird because the `setup.py` has the correct version https://github.com/hyperopt/hyperopt/blob/55bc17fba8204743a7d3d25f779bd8b2e7db70cd/setup.py#L5

maxpumperla:
yeah, I usually make sure to change both, but forgot last time. The following notes should be adapted, too: https://github.com/hyperopt/hyperopt/blob/410d4e4488bb0e5ae51ccc8b0825930d42741b37/RELEASE.txt

marctorsoc:
Personally, I like this solution https://stackoverflow.com/a/7071358

liangz1:
@marctorsoc Feel free to review the PR ^

PR #692, opened 2020-06-27
Unify __version__ in __init__.py and setup.py

liangz1:
This PR fixes issue https://github.com/hyperopt/hyperopt/issues/657.

In future releases, we just need to update the __version__ variable in hyperopt/__init__.py and setup.py will pick up the version number from there. I also updated the RELEASE.txt.


marload:
Looks Good to Me 👍

PR #695, opened 2020-06-30
Add traceback for the error in failed spark trials

lu-wang-dl:
Add the traceback for failed trials in sparkTrials.

### DIFF
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -1,4 +1,4 @@
-ecfrom __future__ import print_function
+from __future__ import print_function
 
 import copy
 import threading
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -476,10 +476,10 @@ class _SparkFMinState:
                 try:
                     result = domain.evaluate(params, ctrl=None, attach_attachments=False)
                     yield result
-                except BaseException:
+                except BaseException as e:
                     error_string = traceback.format_exc()
                     logger.error(error_string)
-                    yield BaseException(error_string)
+                    yield type(e)(error_string)
 
             try:
                 worker_rdd = self.spark.sparkContext.parallelize([0], 1)
--- a/hyperopt/spark.py
+++ b/hyperopt/spark.py
@@ -339,6 +339,10 @@ class _SparkFMinState:
         trial["refresh_time"] = now
         logger.debug("trial task {tid} started".format(tid=trial["tid"]))
 
+    @staticmethod
+    def _get_traceback(err):
+        return err.__dict__.get("_tb_str")
+
     def _finish_trial_run(self, is_success, is_cancelled, trial, data):
         """
         Call this method when a trial evaluation finishes. It will save results to the
@@ -365,8 +369,8 @@ class _SparkFMinState:
             self._write_result_back(trial, result=data)
         else:
             logger.error(
-                "trial task {tid} failed, exception is {e}".format(
-                    tid=trial["tid"], e="".join(data._tb_str)
+                "trial task {tid} failed, exception is {e}.\n {tb}".format(
+                    tid=trial["tid"], e=str(data), tb=self._get_traceback(data)
                 )
             )
             self._write_exception_back(trial, e=data)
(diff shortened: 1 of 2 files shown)


hyperopt/spark.py
```diff
@@ -476,7 +477,9 @@ def run_task_on_executor(_):
                     result = domain.evaluate(params, ctrl=None, attach_attachments=False)
                     yield result
                 except BaseException as e:
-                    yield e
+                    error_string = traceback.format_exc()
+                    logger.error(error_string)
+                    yield e(error_string)
```
liangz1, line 482: I am not familiar with the usage of `e(error_string)`. I tried the following:
```try:
    print(test)
except BaseException as e:
    raise e("add some msg")
```
and got `TypeError: 'NameError' object is not callable`

lu-wang-dl: Yes. It requires to use the Exception class. updated.

hyperopt/spark.py
```diff
@@ -468,15 +468,18 @@ def run_task_thread():
...
 
                 try:
                     result = domain.evaluate(params, ctrl=None, attach_attachments=False)
                     yield result
-                except BaseException as e:
-                    yield e
+                except BaseException:
+                    error_string = traceback.format_exc()
+                    logger.error(error_string)
+                    yield BaseException(error_string)
```
liangz1, line 482: We may want to use `yield type(e)(error_string)` to preserve the error type instead of using `BaseException`.

hyperopt/spark.py
```diff
@@ -476,7 +477,9 @@ def run_task_on_executor(_):
                     result = domain.evaluate(params, ctrl=None, attach_attachments=False)
                     yield result
                 except BaseException as e:
-                    yield e
+                    error_string = traceback.format_exc()
+                    logger.error(error_string)
+                    yield type(e)(error_string)
```
mengxr, line 482: It is not guaranteed that you can reconstruct the error just using an error string. Does `e` already contain the stacktrace or can you use `with_traceback`?

lu-wang-dl: When pass `e` back to driver the traceback info is lost. We have to pass it by string.

mengxr: Did you try `with_traceback`?

mengxr: The issue is that is it not guaranteed that you can reconstruct the error this way. It might be okay to return the error without the traceback.

lu-wang-dl: We have a log as debug for the error message. https://github.com/hyperopt/hyperopt/blob/80b6850214f158ef966011a94bf95969d79e9b0c/hyperopt/spark.py#L367
We can change the level to show it to the users.

lu-wang-dl: @mengxr 
With 
```
tb = e.__traceback__
error_string = traceback.format_exception(type(e), e, tb)
```
Is that guaranteed to get the error message at least to the driver?

marctorsoc:
@lu-wang-dl please fix tests

hyperopt/spark.py
mengxr: the error message is always pickled.

marctorsoc:
> @lu-wang-dl please fix tests

bump

hyperopt/spark.py
```diff
@@ -468,6 +469,7 @@ def run_task_thread():
             params = self._get_spec_from_trial(trial)
 
             def run_task_on_executor(_):
+                import traceback
```
marctorsoc, line 472: please add imports on the top of the file only. Ideally we should use isort and black as well for nicely formatted files

hyperopt/spark.py
```diff
@@ -476,6 +478,10 @@ def run_task_on_executor(_):
                     result = domain.evaluate(params, ctrl=None, attach_attachments=False)
                     yield result
                 except BaseException as e:
+                    tb = e.__traceback__
+                    error_string = traceback.format_exception(type(e), e, tb)
+                    logger.error(error_string)
+                    e._tb_str = error_string
```
marctorsoc, line 484:
```
error_string = traceback.format_exception(type(e), e, tb)
                    logger.error(error_string)
                    e._tb_str = error_string
```
to
```
e._tb_str = traceback.format_exception(type(e), e, tb)
logger.error(e._tb_str)
```
?

lu-wang-dl:
Working on it. Sorry for the delay.

PR #697, opened 2020-07-06
ATPE: set n_features_in_ attribute for scaling model

### LINKED ISSUE #668: ATPE RuntimeError: The reset parameter is False but there is no n_features_in_ attribute
**Version: 0.2.4**
I'm trying to optimise a KNN algorithm using Adaptive TPE. Here's a snippet of the code I'm running
```{python}
KNN_DEFAULT_SPACE = {
    'k': scope.int(hp.quniform('k', 1, 500, 1)),
    'min_k': scope.int(hp.quniform('min_k', 1, 50, 1))
}

def _knn_algo(params):
     # Run KNN with params and crossvalidate to return loss
     return loss

def _objective(params):
    loss = _knn_algo(params)

    return {
        'loss': loss,
        'status': STATUS_OK,
        'hyperparams': params
    }

trials = Trials()

best = fmin(
            objective,
            KNN_DEFAULT_SPACE,
            algo=atpe.suggest,
            max_evals=100,
            trials=trials,
        )
```
The first 10 initialization rounds run fine. However after that I run into this exception

```
 File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/fmin.py", line 482, in fmin
    show_progressbar=show_progressbar,
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/base.py", line 686, in fmin
    show_progressbar=show_progressbar,
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/fmin.py", line 509, in fmin
    rval.exhaust()
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/fmin.py", line 330, in exhaust
    self.run(self.max_evals - n_done, block_until_done=self.asynchronous)
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/fmin.py", line 266, in run
    new_ids, self.domain, trials, self.rstate.randint(2 ** 31 - 1)
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/atpe.py", line 1613, in suggest
    hyperparameterConfig, results, currentTrials=[]
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/hyperopt/atpe.py", line 763, in recommendNextParameters
    transformed = scalingModel.transform([[stats[feature]]])[0][0]
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/sklearn/preprocessing/_data.py", line 794, in transform
    force_all_finite='allow-nan')
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/sklearn/base.py", line 436, in _validate_data
    self._check_n_features(X, reset=reset)
  File "/home/rohan/auto-surprise-venv/lib/python3.6/site-packages/sklearn/base.py", line 373, in _check_n_features
    "The reset parameter is False but there is no "
RuntimeError: The reset parameter is False but there is no n_features_in_ attribute. Is this estimator fitted?
```
It seems that the issue arises when transforming the scaling model. Any ideas on the exact reason for this?

Thanks in advance

thededlier:
From sklearn 0.23.0, the n_features_in_ attribute is verified when performing fit or transform in StandardScaler. 

This will fix #668

### DIFF
--- a/hyperopt/tests/test_atpe_basic.py
+++ b/hyperopt/tests/test_atpe_basic.py
@@ -25,7 +25,8 @@ class TestATPE(unittest.TestCase):
         # minimize the objective over the space
         from hyperopt import fmin, atpe, space_eval
 
-        best = fmin(objective, space, algo=atpe.suggest, max_evals=10)
+        # Max evals should be greater than 10, as the first 10 runs are only the initialization rounds
+        best = fmin(objective, space, algo=atpe.suggest, max_evals=20)
 
         print(best)
         # -> {'a': 1, 'c2': 0.01420615366247227}
--- a/hyperopt/tests/test_atpe_basic.py
+++ b/hyperopt/tests/test_atpe_basic.py
@@ -1,19 +1,17 @@
 from __future__ import absolute_import
-import unittest
+from hyperopt import hp, fmin, atpe, space_eval
 
 
-class TestATPE(unittest.TestCase):
+class TestATPE:
     def test_run_basic_search(self):
         def objective(args):
             case, val = args
-            if case == "case 1":
+            if case == "case 2":
                 return val
             else:
                 return val ** 2
 
         # define a search space
-        from hyperopt import hp
-
         space = hp.choice(
             "a",
             [
@@ -23,12 +21,9 @@ class TestATPE(unittest.TestCase):
         )
 
         # minimize the objective over the space
-        from hyperopt import fmin, atpe, space_eval
-
-        # Max evals should be greater than 10, as the first 10 runs are only the initialization rounds
+        # NOTE: Max evals should be greater than 10, as the first 10 runs are only the initialization rounds
         best = fmin(objective, space, algo=atpe.suggest, max_evals=20)
 
-        print(best)
-        # -> {'a': 1, 'c2': 0.01420615366247227}
-        print(space_eval(space, best))
-        # -> ('case 2', 0.01420615366247227}
+        # Assert that case 2 was the best choice
+        assert best["a"] == 1
+        assert space_eval(space, best)[0] == "case 2"
--- a/hyperopt/tests/test_atpe_basic.py
+++ b/hyperopt/tests/test_atpe_basic.py
@@ -1,29 +1,38 @@
 from __future__ import absolute_import
+import random
+import numpy as np
 from hyperopt import hp, fmin, atpe, space_eval
 
+random.seed(1)
+np.random.seed(1)
 
-class TestATPE:
-    def test_run_basic_search(self):
-        def objective(args):
-            case, val = args
-            if case == "case 2":
-                return val
-            else:
-                return val ** 2
 
-        # define a search space
-        space = hp.choice(
-            "a",
-            [
-                ("case 1", 1 + hp.lognormal("c1", 0, 1)),
-                ("case 2", hp.uniform("c2", -10, 10)),
-            ],
-        )
+def test_run_basic_search():
+    def objective(args):
+        case, val = args
+        if case == "case 1":
+            return val
+        else:
+            return val ** 2
 
-        # minimize the objective over the space
-        # NOTE: Max evals should be greater than 10, as the first 10 runs are only the initialization rounds
-        best = fmin(objective, space, algo=atpe.suggest, max_evals=20)
+    # define a search space
+    space = hp.choice(
+        "a",
+        [
+            ("case 1", 1 + hp.lognormal("c1", 0, 1)),
+            ("case 2", hp.uniform("c2", -10, 10)),
+        ],
+    )
 
-        # Assert that case 2 was the best choice
-        assert best["a"] == 1
-        assert space_eval(space, best)[0] == "case 2"
+    # minimize the objective over the space
+    # NOTE: Max evals should be greater than 10, as the first 10 runs are only the initialization rounds
+    best = fmin(
+        objective,
+        space,
+        algo=atpe.suggest,
+        max_evals=20,
+        rstate=np.random.RandomState(1),
+    )
+
+    assert best["a"] == 1
+    assert space_eval(space, best)[0] == "case 2"
--- a/hyperopt/atpe.py
+++ b/hyperopt/atpe.py
@@ -163,7 +163,6 @@ class Hyperparameter:
             v = "anyOf" if "anyOf" in self.config else "oneOf"
             data = self.config[v]
 
-
             for index, param in enumerate(data):
                 subKeys = Hyperparameter(
                     param, self, name + "." + str(index)
@@ -670,6 +669,7 @@ class ATPEOptimizer:
             self.featureScalingModels[key].var_ = numpy.array(
                 scalingModelData[key]["variances"]
             )
+            self.featureScalingModels[key].n_features_in_ = 1
 
         self.parameterModels = {}
         self.parameterModelConfigurations = {}


JSunRae:
I have this problem with conda scikit-learn 0.23.1

marctorsoc:
could you add a test that fails without this? see https://github.com/hyperopt/hyperopt/issues/565 to find an example

thededlier:
@marctorsoc I've updated the sanity test so that it covers rounds after initialisation (where the error would occur). 

Just FYI this does not fix the `IndexError` mentioned in #565

hyperopt/tests/test_atpe_basic.py
```diff
@@ -25,7 +25,8 @@ def objective(args):
         # minimize the objective over the space
         from hyperopt import fmin, atpe, space_eval
 
-        best = fmin(objective, space, algo=atpe.suggest, max_evals=10)
+        # Max evals should be greater than 10, as the first 10 runs are only the initialization rounds
+        best = fmin(objective, space, algo=atpe.suggest, max_evals=20)
```
marctorsoc, line 29: The first thing i thought when I saw your change was: 

"maybe rather than increasing the max_evals to 20, you can reduce the init rounds to e.g. 2-3?" 

however, you're editing another test, which is supposed to test something. I know it's not explicitly part of your task, but could you turn the 

```
print(best)
        # -> {'a': 1, 'c2': 0.01420615366247227}
        print(space_eval(space, best))
        # -> ('case 2', 0.01420615366247227}
```
into actual checks with unittest assert? and please move imports to the top and run isort, black

I hope this shouldn't take you more than 10min and would increase the hygiene of this file. I would also ask you to turn the test into pytest but appreciate that's too much to ask. If you need a hand, please let me know!

thededlier: Updated the test. I had to change the objective function so that its more consistently returns case 2.

thededlier:
> LGTM! just added some minor comments. I'm conscious this won't fix the `IndexError` but I expect it to fix the
> 
> ```
> RuntimeError: The reset parameter is False but there is no n_features_in_ attribute. Is this estimator fitted?
> ```
> 
> right?

Yup, that's right

hyperopt/tests/test_atpe_basic.py
marctorsoc: some comments:

- what do you mean it's more consistently ...? it's not ALWAYS returning case2? if so, please set the random seed to ensure determinism
- could you reduce the initial rounds to e.g. 5, so max_evals can stay the same?
- you can remove the class as well and just leave `test_run_basic_search` at the top level

hyperopt/tests/test_atpe_basic.py
```diff
@@ -1,19 +1,17 @@
...
+from hyperopt import hp, fmin, atpe, space_eval
 
 
-class TestATPE(unittest.TestCase):
+class TestATPE:
     def test_run_basic_search(self):
         def objective(args):
             case, val = args
-            if case == "case 1":
+            if case == "case 2":
```
marctorsoc, line 9: why do you need this change? fwiw we don't need to have a realistic objective function here, just check that the end2end works

thededlier: Alright, so just `assert best` ?

hyperopt/tests/test_atpe_basic.py
thededlier: - Yes, it was not always returning case2. But if we are only checking if end2end works then I guess this can be ignored and I'll revert it.

- I can't really change the initial rounds in a test case without altering the code for ATPE, which I think would be better left avoided. See https://github.com/hyperopt/hyperopt/blob/8edf67974e68a01ed553f768166bb83812ac0578/hyperopt/atpe.py#L715

- Alright will do

marctorsoc: Your approach is perfect with those asserts. I just meant to have something that's deterministic, even if practically speaking is not a fancy objective function doing something realistic. Forget about changing the objective function, just set the random seed so that it always returns the same, and it's good to go! Sorry for the confusion

hyperopt/tests/test_atpe_basic.py
marctorsoc: see comment above

hyperopt/tests/test_atpe_basic.py
thededlier: Ok thanks for clearing that up. I'm still having some issues setting the random state. I'm setting `rstate=np.random.RandomState(1)` however the results still change. I tried the same with `tpe.suggest` and it seems to work fine. Am I missing something here?

marctorsoc: https://stackoverflow.com/a/22994890 ?

marctorsoc: might also be that the `random` python library is used, so you need to fix that seed as well? This is definitely something to review, we should set the global random seeds before running tests, not for each test separately (but not part of your PR ofc)

thededlier: yeah that seems to be the case. I Needed to set both `random.seed` and `numpy.random.seed`. 

https://github.com/hyperopt/hyperopt/blob/8edf67974e68a01ed553f768166bb83812ac0578/hyperopt/atpe.py#L698

sunshinejnjn:
Just want to provide a simple solution here, **for the StandardScaler only**:

1. pickle load the scaler (dumped with sklearn version 0.22-) with sklearn version 0.23 as, lets say, 
`scaler1=pickle.load(somefile)`
 . ignore the warning here.
2. scaler1.n_features_in_=**THE_NUMBER_OF_FEATURES_OF_THE_SCALER**, something like 
`scaler1.n_features_in_=**1**`
3. pickle dump (or libwork) it back to the file.
`pickle.dump(scaler1, somefile)`

After that, you'll be fine with it, I hope.

marctorsoc:
> Just want to provide a simple solution here, **for the StandardScaler only**:
> 
> 1. pickle (or joblib) load the scaler (dumped with sklearn version 0.22-) with sklearn version 0.23 as, lets say,
>    `scaler1=pickle.load(somefile)`
>    . ignore the warning here.
> 2. scaler1.n_features_in_=**THE_NUMBER_OF_FEATURES_OF_THE_SCALER**, something like
>    `scaler1.n_features_in_=1`
> 3. pickle dump (or joblib) it back to the file.
>    `pickle.dump(scaler1, somefile)`
> 
> After that, you'll be fine with it, I hope.
> 
> BTW: If it is a OneHotEncoder, you probably want to set scaler1.drop_idx_=None by the way.

not sure I understand. Do you mean to solve the `IndexError` issue?

sunshinejnjn:
> > Just want to provide a simple solution here, **for the StandardScaler only**:
> > 
> > 1. pickle (or joblib) load the scaler (dumped with sklearn version 0.22-) with sklearn version 0.23 as, lets say,
> >    `scaler1=pickle.load(somefile)`
> >    . ignore the warning here.
> > 2. scaler1.n_features_in_=**THE_NUMBER_OF_FEATURES_OF_THE_SCALER**, something like
> >    `scaler1.n_features_in_=1`
> > 3. pickle dump (or joblib) it back to the file.
> >    `pickle.dump(scaler1, somefile)`
> > 
> > After that, you'll be fine with it, I hope.
> > BTW: If it is a OneHotEncoder, you probably want to set scaler1.drop_idx_=None by the way.
> 
> not sure I understand. Do you mean to solve the `IndexError` issue?

The .drop_idx_=None would solve the "IndexError" with some encoders (such as OneHotEncoder).
But the first part was primarily for the "RuntimeError: The reset parameter is False but there is no n_features_in_ attribute. Is this estimator fitted?" error.

PR #700, opened 2020-07-09
Use ==/!= to compare constant literals (str, bytes, int, float, tuple)

cclauss:
Avoid [SyntaxWarnings on Python >= 3.8](https://docs.python.org/3/whatsnew/3.8.html#porting-to-python-3-8).

% `python3.8`
```
>>> "" is ""
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```

### DIFF
--- a/hyperopt/atpe.py
+++ b/hyperopt/atpe.py
@@ -1528,7 +1528,7 @@ class ATPEOptimizer:
 
             for param in Hyperparameter(hyperparameterSpace).getFlatParameters():
                 value = result[param.name]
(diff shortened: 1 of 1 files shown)


maxpumperla:
@cclauss thanks for this PR! (I'm slightly mad that this is discouraged now, because I think it makes for very readable, pythonic syntax)

cclauss:
;-). It does look good but can lead to hard-to-spot bugs where equality is not the same thing as identity in Python.

PR #702, opened 2020-07-09
Travis CI: Add Python 3.8 to tests in allow_failures mode

cclauss:
[Ubuntu Trusty is End of Life](https://releases.ubuntu.com) but when we try to upgrade to Xenial our Travis CI tests fail.

This PR creates adds Python 3.6 on Xenial run in __allow_failures__ mode so we can work to make the code compatible without breaking the build.  Once the tests pass on Xenial, we can remove allow_failures and add Python 3.7 and 3.8 to the testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: linux
-dist: xenial
+dist: trusty
 language: python
 
 cache:
@@ -26,17 +26,22 @@ jobs:
     - python: 3.6
       env:
         - IPYTHON=ipython[all]
-    - python: 3.7
+    - python: 3.6
+      dist: xenial
       env:
         - IPYTHON=ipython[all]
+    #- python: 3.7
+    #  dist: xenial
+    #  env:
+    #    - IPYTHON=ipython[all]
+    #    - NUMPY_VERSION=1.14.6
+    #- python: 3.8
+    #  dist: xenial
+    #  env:
+    #    - IPYTHON=ipython[all]
     #    - NUMPY_VERSION=1.19.0
-    - python: 3.8
-      env:
-        - IPYTHON=ipython[all]
-        - NUMPY_VERSION=1.19.0
   allow_failures:
-    - python: 3.7
-    - python: 3.8
+    - dist: xenial
 
 services:
   - mongodb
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -26,6 +26,22 @@ matrix:
...
+    #- python: 3.7
+    #  dist: xenial
+    #  env:
+    #    - IPYTHON=ipython[all]
+    #    - NUMPY_VERSION=1.14.6
+    #- python: 3.8
+    #  dist: xenial
+    #  env:
+    #    - IPYTHON=ipython[all]
+    #    - NUMPY_VERSION=1.19.0
```
marctorsoc, line 42: please remove commented code if not needed

.travis.yml
```diff
@@ -48,8 +64,7 @@ install:
   - pip install .[SparkTrials]
   - pip install .[ATPE]
   - pip install pytest>=3.6
-  - pip install pytest-cov pep8 pytest-pep8
-  - pip install wrapt
+  - pip install pytest-cov pep8 pytest-pep8 wrapt
```
marctorsoc, line 67: I wonder whether we could use `pyproject.toml` , see `https://www.python.org/dev/peps/pep-0518/#specification`, and forget about all these hardcoded deps @maxpumperla

.travis.yml
cclauss: We are going to have to run on Python 3.7 and 3.8 to be credible.

However, given that we cannot yet figure out how to run on Python 3.6 on Xenial, let’s keep Python 3.7 and 3.8 commented out until we figure that out.

marctorsoc: WDYT @maxpumperla @marload ?

marctorsoc: I'm not an expert on this, but could we change to another dist that works for 3.7 and 3.8? @cclauss

marctorsoc: bump @marload @maxpumperla @cclauss

maxpumperla:
@cclauss any update here? It's a good contribution, but we have to resolve the issues. If we don't have a solution, it might not make sense to keep this PR open.

maxpumperla:
I just realised that master is currently better off due to the other Spark PR. @cclauss closing this for now, let me know if you want to add anything.

PR #713, opened 2020-07-29
Refactor/string concat

marload:
It is better to use join than += when concating multiple strings.

Thank You!


docs/autogen.py
```diff
@@ -53,19 +53,15 @@ def get_function_signature(function, method=True):
...
-            v = "'" + v + "'"
-        st += str(a) + "=" + str(v) + ", "
-    if kwargs or args:
-        signature = st[:-2] + ")"
-    else:
-        signature = st + ")"
-    return signature
+            v = "'{}'".format(v)
+        st.append("{}={}".format(a, v))
+    st = ', '.join(st) + ")"
```
marctorsoc, line 64: you're not returning anything. If this didn't fail, then maybe we should have a test to make sure this should have failed

docs/autogen.py
```diff
@@ -53,19 +53,15 @@ def get_function_signature(function, method=True):
         args = args[: -len(defaults)]
     else:
         kwargs = []
-    st = "%s.%s(" % (clean_module_name(function.__module__), function.__name__)
+    st = ["%s.%s(" % (clean_module_name(function.__module__), function.__name__)]
```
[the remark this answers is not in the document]
marctorsoc: until that happens, let's use the `.format` format? it's easier to read and would be homogeneous with the code later

docs/autogen.py
```diff
@@ -53,19 +53,15 @@ def get_function_signature(function, method=True):
...
 
     for a in args:
-        st += str(a) + ", "
+        st.append(str(a))
     for a, v in kwargs:
```
marctorsoc, line 60: maybe a bit more intuitive names than `a`, `v`  and `st`?

docs/autogen.py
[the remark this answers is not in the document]
marload: The f-string has not been used because python 2.7 has not been completely removed yet. But I think 2.7 should be removed in a short time. 😄

docs/autogen.py
marload: complete 👍

docs/autogen.py
marload: I change `a` to `arg`, `v` to `value` and `st` to `signature` 😃

marctorsoc: I guess the `a` in kwargs should be `key`?

marload:
@marctorsoc 
I will add the test code ASAP. We'd better merge after that. :+1:

marctorsoc:
how are you doing with this @marload ?

maxpumperla:
@marload just pinging to see if you might have time to look at this again.

marload:
Hmm... Where should I make the test code?
`autogen.'py' is just a tool script to create a document, not part of the library. So it doesn't seem good to put the test code for this in 'hyperopt/tests'. What do you think? @maxpumperla @marctorsoc

maxpumperla:
@marload agreed, just a script. If you decide to test, put the test right next to the script. But we should definitely do a manual check to confirm that the output from `mkdocs` is consistent with what we had before.

docs/autogen.py
```diff
@@ -53,19 +53,15 @@ def get_function_signature(function, method=True):
...
-    return signature
+    signature = ["{}.{}(".format(clean_module_name(function.__module__), function.__name__)]
+
+    for arg in args:
+        signature.append(str(arg))
+    for key, value in kwargs:
+        if isinstance(value, str):
+            value = "'{}'".format(value)
+        signature.append("{}={}".format(key, value))
+    return ', '.join(signature) + ")"
```
marctorsoc, line 64: given that we finally released the last python2 version, please use fstrings in these changes

e.g. `value = f"'{value}'"

marctorsoc:
any chance you can finish this @marload ? it looks really close to the finish line :)

marload:
@marctorsoc 
I still don't know how to make a test code for `autogen.py`. This PR has been open for too long, so it would be better to merge first and create a testcode later.

What do you think?

marctorsoc (approved):

PR #737, opened 2020-09-10
Add fmin optional arg pickle_return (if return_argmin is True)

brethvoice:
Allows user to specify a name for the returned dictionary.  Just in case some code in the calling script for fmin breaks, at least the hours of compute time will not be wasted.

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -10,6 +10,7 @@ import os
 import sys
 import time
 from timeit import default_timer as timer
+import pickle
 
 import numpy as np
 
@@ -372,6 +373,7 @@ def fmin(
     max_queue_len=1,
     show_progressbar=True,
     early_stop_fn=None,
+    pickle_return=None,
 ):
     """Minimize a function over a hyperparameter space.
 
@@ -469,10 +471,13 @@ def fmin(
     show_progressbar : bool or context manager, default True (or False is verbose is False).
         Show a progressbar. See `hyperopt.progress` for customizing progress reporting.
 
-    early_stop_fn: callable ((result, *args) -> (Boolean, *args)).
+    early_stop_fn : callable ((result, *args) -> (Boolean, *args)).
         Called after every run with the result of the run and the values returned by the function previously.
         Stop the search if the function return true.
         Default None.
+
+    pickle_return : str, default None
+        Specifies prefix of .pkl file to save returned argmin dictionary.
 
     Returns
     -------
@@ -543,6 +548,9 @@ def fmin(
             raise Exception(
                 "There are no evaluation tasks, cannot return argmin of task losses."
             )
+        if pickle_return is not None:
+            with open(pickle_return + '.pkl', 'wb') as out_file:
+                pickle.dump(trials.argmin, out_file)
         return trials.argmin
     if len(trials) > 0:
         # Only if there are some successful trail runs, return the best point in
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -549,7 +549,7 @@ def fmin(
                 "There are no evaluation tasks, cannot return argmin of task losses."
             )
         if pickle_return is not None:
-            with open(pickle_return + '.pkl', 'wb') as out_file:
+            with open(str(pickle_return) + '.pkl', 'wb') as out:file:
                 pickle.dump(trials.argmin, out_file)
         return trials.argmin
     if len(trials) > 0:


brethvoice:
Sorry for the confusion in the original pull request commit.  The pickle_return is used only if return_argmin is True (not False).

hyperopt/fmin.py
```diff
@@ -469,11 +471,14 @@ def fmin(
...
         Show a progressbar. See `hyperopt.progress` for customizing progress reporting.
 
-    early_stop_fn: callable ((result, *args) -> (Boolean, *args)).
+    early_stop_fn : callable ((result, *args) -> (Boolean, *args)).
         Called after every run with the result of the run and the values returned by the function previously.
         Stop the search if the function return true.
         Default None.
 
+    pickle_return : str, default None
+        Specifies prefix of .pkl file to save returned argmin dictionary.
```
marctorsoc, line 480: what's the behaviour if None?

hyperopt/fmin.py
```diff
@@ -543,6 +548,9 @@ def fmin(
             raise Exception(
                 "There are no evaluation tasks, cannot return argmin of task losses."
             )
+        if pickle_return is not None:
+            with open(pickle_return + '.pkl', 'wb') as out_file:
```
marctorsoc, line 552: please black this

marctorsoc (changes requested):

hyperopt/fmin.py
brethvoice: Nothing happens if None.

hyperopt/fmin.py
brethvoice: Could you change this to force `pickle_return` to be a string to avoid errors.  So line 552 would be:
```suggestion
            with open(str(pickle_return) + '.pkl', 'wb') as out:file:
```
I used the black [playground](https://black.now.sh) and my stuff came back the same way I put it in.  I have never used black before.

hyperopt/fmin.py
```diff
@@ -549,7 +549,7 @@ def fmin(
                 "There are no evaluation tasks, cannot return argmin of task losses."
             )
         if pickle_return is not None:
```
marctorsoc, line 551: if isinstance(pickle_return, str)

hyperopt/fmin.py
```diff
@@ -549,7 +549,7 @@ def fmin(
...
-            with open(pickle_return + '.pkl', 'wb') as out_file:
+            with open(str(pickle_return) + '.pkl', 'wb') as out:file:
```
marctorsoc, line 552: go back to previous and do double quotes

bjkomer:
It looks to me like this change will make:

```python
best = fmin(...)
pickle.dump(best, open('fname.pkl', 'wb'))
```
be the same as:

```python
best = fmin(..., pickle_return='fname')
```

I don't think that alone is worth adding a new parameter to `fmin`, as the user has to do about the same amount of work in both cases and if there is an error during the `fmin` call the data is lost either way.

That being said, I think it would be extremely useful to have a simple parameter like this to specify a file to save all of the trials data to periodically, so if `fmin` is interrupted, no data is lost.

marctorsoc:
> It looks to me like this change will make:
> 
> ```python
> best = fmin(...)
> pickle.dump(best, open('fname.pkl', 'wb'))
> ```
> 
> be the same as:
> 
> ```python
> best = fmin(..., pickle_return='fname')
> ```
> 
> I don't think that alone is worth adding a new parameter to `fmin`, as the user has to do about the same amount of work in both cases and if there is an error during the `fmin` call the data is lost either way.
> 
> That being said, I think it would be extremely useful to have a simple parameter like this to specify a file to save all of the trials data to periodically, so if `fmin` is interrupted, no data is lost.

I agree with you, any argument to defend your PR @brethvoice ? otherwise will close it

About dumping trials periodically, sounds like a good idea to avoid losing the progress. It'd be great being able to load and resume. If you want to contribute to this @bjkomer , I'm very happy to review a PR :)

brethvoice:
Good review and comments.  Thanks for consideration and I respect the decision to reject my request.

hyperopt/fmin.py
brethvoice: Pull request closed/withdrawn.  Good to know about double quotes...I thought single quotes were the standard.

marctorsoc:
Soon we'll only support python3.7+ and then black will be a check in PRs, so your code would have failed. Can't wait for that :)

PR #742, opened 2020-09-18
Drop Python < 3.7 from CI

marctorsoc:
Initial step towards release 1.0, supporting python>=3.7 only

Release 0.2.5 is the last version supporting python < 3.7


maxpumperla:
@marctorsoc cool, that's moving in the right direction! do we really want to bump to 1.0? all things considered, this is a very minor change. maybe just 0.3?

If and when we come up with a roadmap that includes some major (potentially API breaking) changes, maybe we can plan for a 1.0 release.

maxpumperla:
p.s. can you please fix the conflicts?

marctorsoc:
> @marctorsoc cool, that's moving in the right direction! do we really want to bump to 1.0? all things considered, this is a very minor change. maybe just 0.3?
> 
> If and when we come up with a roadmap that includes some major (potentially API breaking) changes, maybe we can plan for a 1.0 release.

ok, I will remove the release version change (which is not really doing anything as we won't release yet), but once we (for example) add black, that doesn't support Python2 so it's a breaking change and should be released as 1.0, IMHO

marctorsoc:
> p.s. can you please fix the conflicts?

yup, will do

marctorsoc:
can anyone approve? @maxpumperla @marload @jaberg

maxpumperla:
@marctorsoc great to see this simplification! thanks

PR #743, opened 2020-09-21
option to automatically save trials to a file each iteration

bjkomer:
The user can give a filename to `fmin` where the `Trials` object will be saved to each iteration using pickle. This allows the optimization to be easily continued from where it left off if interrupted. If no trials argument is given to `fmin` it will load from this file if it exists instead of creating a new empty `Trials`. If a trials argument is given, it will use that one instead. Either way it will overwrite the file given to `trials_save_file` each iteration with the most up to date trials object it is using.

Simple example of usage:

```python
from hyperopt import Trials, tpe, hp, fmin

def objective(x):
    return {'loss':x**2,'status':hyperopt.STATUS_OK}

space = hp.uniform('x', -10, 10)

trials_fname = 'my_trials.pkl'

best = fmin(fn=objective, space=space, algo=tpe.suggest, trials_save_file=trials_fname, max_evals=100)
```

### DIFF
--- a/hyperopt/base.py
+++ b/hyperopt/base.py
@@ -654,6 +654,7 @@ class Trials(object):
         return_argmin=True,
         show_progressbar=True,
         early_stop_fn=None,
+        trials_save_file="",
     ):
         """Minimize a function over a hyperparameter space.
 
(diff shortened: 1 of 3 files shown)


maxpumperla:
that's super cool! thanks

PR #747, opened 2020-09-25
Fix comparison glitch

magdapoppins:
`is not` should not be used for comparing strings, as it is an identity check and not an equality check.


magdapoppins:
Oops, just noticed this was already fixed. Will fetch again.

PR #748, opened 2020-09-25
Cleanup

magdapoppins:
- Run formatter
- (Suggestion) Add github action to automatically lint on push

PR #751, opened 2020-10-08
GitHub Action to lint Python code

cclauss:
Output: https://github.com/cclauss/hyperopt/actions 

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead, these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python, they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python, a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -7,6 +7,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
       - run: pip install black==20.8b1 codespell flake8
-      - run: black --check . || true
+      - run: black --check .
       - run: codespell --ignore-words-list="hist,optionn" --quiet-level=2  # --skip=""
       - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,12 @@
+name: lint_python
+on: push
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install black==20.8b1 codespell flake8
+      - run: black --check . || true
+      - run: codespell --quiet-level=2  # --ignore-words-list="" --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -8,5 +8,5 @@ jobs:
       - uses: actions/setup-python@v2
       - run: pip install black==20.8b1 codespell flake8
       - run: black --check . || true
-      - run: codespell --quiet-level=2  # --ignore-words-list="" --skip=""
+      - run: codespell --ignore-words-list="hist,optionn" --quiet-level=2  # --skip=""
       - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
--- a/docs/autogen.py
+++ b/docs/autogen.py
@@ -10,15 +10,7 @@ import inspect
 import os
 import shutil
 
-import sys
-
-if sys.version[0] == "2":
-    reload(sys)
-    sys.setdefaultencoding("utf8")
-
-
 EXCLUDE = {}
-
 
 PAGES = [
     # {
--- a/hyperopt/pyll/base.py
+++ b/hyperopt/pyll/base.py
@@ -364,7 +364,7 @@ class Apply(object):
         if extra_args_ok:
             # XXX: THIS IS NOT BEING TESTED AND IS OBVIOUSLY BROKEN
             # TODO: 'args' does not even exist at this point
-            binding[args_param].extend(args[code.co_argcount :])
+            binding[args_param].extend(args[code.co_argcount :])  # noqa: F821
 
         # -- bind keyword arguments
         for aname, aval in self.named_args:
@@ -729,8 +729,8 @@ def clone_merge(expr, memo=None, merge_literals=False):
     #    XXX node.arg does not always work (builtins, weird co_flags)
     node_args = [(node.pos_args, node.named_args) for node in nodes]
     try:
-        del node
-    except:
+        del node  # noqa: F821
+    except NameError:
         pass
     for ii, node_ii in enumerate(nodes):
         if node_ii in memo:
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -490,8 +490,8 @@ def fmin(
     argmin : dictionary
         If return_argmin is True returns `trials.argmin` which is a dictionary.  Otherwise
         this function  returns the result of `hyperopt.space_eval(space, trails.argmin)` if there
-        were successfull trails. This object shares the same structure as the space passed.
-        If there were no successfull trails, it returns None.
+        were successful trails. This object shares the same structure as the space passed.
+        If there were no successful trails, it returns None.
     """
     if algo is None:
         algo = tpe.suggest


.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,29 @@
...
+          python-version: ${{ matrix.python-version }}
+      - uses: DigiPie/mongo-action@v1
+      - run: pip install black codespell flake8 isort pytest
+      - run: black --check . || true
+      # - run: black --diff . || true
+      # - if: matrix.python-version >= 3.6
+      #  run: |
+      #    pip install black
+      #    black --check . || true
+      - run: codespell --ignore-words-list="hist,optionn" --quiet-level=2 || true  # --skip=""
```
marctorsoc, line 24: maybe add a comment here to explain what the quiet-level is for? my understand is that the `|| true` is disabling this. Not sure I see the point on adding the line if it's always passing. How noisy is this at the moment?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,29 @@
...
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --profile black . || true
```
marctorsoc, line 26: I'm happy with applying isort to the whole codebase and then enabling this

marctorsoc: we should then add isort to the README, as a requirement

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,29 @@
...
+      - run: pip install -r requirements.txt || true
+      - run: pip install bson future networkx nose numpy pyspark scipy tqdm
+      - run: pytest . || true
```
marctorsoc, line 29: again, if we add this is without the || true. Otherwise I don't see the point

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,29 @@
...
+      - run: black --check . || true
+      # - run: black --diff . || true
+      # - if: matrix.python-version >= 3.6
+      #  run: |
+      #    pip install black
+      #    black --check . || true
+      - run: codespell --ignore-words-list="hist,optionn" --quiet-level=2 || true  # --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --profile black . || true
+      - run: pip install -r requirements.txt || true
```
marctorsoc, line 27: let's go one by one. This PR is about linting, so let's install as usual `pip install -e '.[extra1, extra2, etc]'` for now (i.e. same commands as we do for travis today)

docs/autogen.py
```diff
@@ -10,16 +10,8 @@
 import os
 import shutil
 
-import sys
-
-if sys.version[0] == "2":
-    reload(sys)
-    sys.setdefaultencoding("utf8")
-
-
```
marctorsoc, line 19: 👏

hyperopt/pyll/base.py
```diff
@@ -368,7 +368,7 @@ def arg(self):
         if extra_args_ok:
             # XXX: THIS IS NOT BEING TESTED AND IS OBVIOUSLY BROKEN
             # TODO: 'args' does not even exist at this point
-            binding[args_param].extend(args[code.co_argcount :])
+            binding[args_param].extend(args[code.co_argcount :])  # noqa: F821
```
marctorsoc, line 371: why you need the noqa here?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,29 @@
+name: lint_python
+on: [pull_request, push]
```
marctorsoc, line 2: I don't think we need the push. If something has been approved and merged, it's because it passes tests. And we won't push to master directly. Am I understanding this correctly?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  lint_python:
+    strategy:
+      fail-fast: false
```
marctorsoc, line 6: I don't think fail-fast should be False. (One of the main) The point of linting is to avoid running the tests if there are major errors that can be found just by code inspection

marctorsoc (changes requested):
Nice work! Thanks for your contribution @cclauss. Added some comments, ideas, and things I think it's better to be conservative

hyperopt/pyll/base.py
cclauss: Because like the comment on the lines above say, `args` is not defined.

marctorsoc: I guess this is working because some black magic with the pyll? or is it directly wrong?

cclauss: I have never heard of pyll so I could not say.  I am not a fan of magic in software engineering.

.github/workflows/lint_python.yml
cclauss: If push is not included the no test is run for the contributor so let’s only run on push.

.github/workflows/lint_python.yml
cclauss: This is useful when testing across Linux, macOS, and Windows and/or multiple versions of Python (the comments that we deleted) as if all a Windows test fails, we can still see if macOS and Linux pass.  Give that we only have one run, fast-fail is a noop.

marctorsoc: I see, I thought it was for the case when running multiple jobs in parallel (e.g. linting and testing)

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,11 @@
+name: lint_python
+on: push
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install black flake8
+      - run: black --check . || true
```
marctorsoc, line 10: isn't this || true disabling the check?

hyperopt/pyll/base.py
marctorsoc: nvm

.github/workflows/lint_python.yml
marctorsoc: can you elaborate more your answer? we want the GHA to be checks in the PRs, and we don't need them when we push to master

.github/workflows/lint_python.yml
cclauss: No.  The check is still being run:
```

Run black --check . || true
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/anneal.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/base.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/fmin.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/pyll/base.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/mongoexp.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/pyll_utils.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/rdists.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/tests/test_domains.py
would reformat /home/runner/work/hyperopt/hyperopt/hyperopt/tpe.py
Oh no! 💥 💔 💥
9 files would be reformatted, 46 files would be left unchanged.
```
It is just not failing the job if black fails.  This is like how [`allow_failures` works in Travis CI](https://docs.travis-ci.com/user/customizing-the-build/#jobs-that-are-allowed-to-fail).  "_This lets you add in experimental and preparatory builds, for example to test against runtime versions or configurations that you are not ready to officially support._"

marctorsoc: yes, sorry, this is what I meant. We do want it to fail, why would we run it otherwise?

.github/workflows/lint_python.yml
cclauss: Pull Request only happens on the main repo end of things.
Push happens both on the contributor's repo and on the main repo.
This means that if only `pull_request` is set then no tests are run when a contributor pushes to a new branch.

.github/workflows/lint_python.yml
cclauss: "to test against runtime versions or configurations that you are not ready to officially support"

Like running codespell in allow_failures mode until #755 lands and then removing to `|| true` to make the test mandatory after #755 lands.  `|| true` is a way to highlight the work that is still to be done.

.github/workflows/lint_python.yml
marctorsoc: fair enough. Would `push` allow to have the tests visible in the PR?

.github/workflows/lint_python.yml
marctorsoc: I see. I thought after https://github.com/hyperopt/hyperopt/pull/748, black was already applied to the whole codebase, so we wouldn't need the `|| true`. But apparently that's not the case? Can we pin a specific version of black (same for travis and GHA)? I don't think it's a good to have this bothering because a new version is released

cclauss: pip install black=20.8b1

cclauss:
Closing in favor of #760

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,12 @@
...
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install black==20.8b1 codespell flake8
+      - run: black --check .
+      - run: codespell --ignore-words-list="hist,optionn" --quiet-level=2  # --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
cclauss, line 12: @marload Two more tests to consider adding to the pre-commit

PR #754, opened 2020-10-08
Change the default value of algo to TPE

marload:
The default algorithm of `fmin` has been changed to TPE.


maxpumperla:
thanks @marload

hyperopt/fmin.py
```diff
@@ -493,6 +494,13 @@ def fmin(
         were succesfull trails. This object shares the same structure as the space passed.
         If there were no succesfull trails, it returns None.
     """
+    if algo is None:
```
marctorsoc, line 497: it would have been good to write this to the docstring. Plus, now that we assume only 3.7+, we could start typehinting :)  let's think about it for next PRs

PR #758, opened 2020-10-24
Add PyUpgrade Pre-Commit Hook for support python 3.7 and above

marload:
Removed format method and added f-string.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,5 @@ repos:
     hooks:
     - id: black
       args: [--check]
+    - id: pyupgrade
+      args: [--py37-plus]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,5 +4,8 @@ repos:
     hooks:
     - id: black
       args: [--check]
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v2.7.3
+    hooks:
     - id: pyupgrade
       args: [--py37-plus]
--- a/docs/autogen.py
+++ b/docs/autogen.py
@@ -7,12 +7,6 @@ import inspect
 import os
 import shutil
 
-import sys
-
-if sys.version[0] == "2":
-    reload(sys)
-    sys.setdefaultencoding("utf8")
-
 
 EXCLUDE = {}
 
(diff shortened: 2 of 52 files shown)


marctorsoc (approved):
Approved, thanks! if you can, could update the ones using % so we have everything with the same format? :)

cclauss:
This process can largely be automated using [pyupgrade](https://github.com/asottile/pyupgrade) but we will need to drop [support for legacy Python](https://github.com/hyperopt/hyperopt/blob/master/setup.py#L33) to use f-strings because they are only available on Python >= 3.6.

marload:
@cclauss 
We internally decided to support HyperOpt in Python 3.7 and above. Using a pyupgrade looks good. :smile:

marload:
@marctorsoc @maxpumperla @cclauss 
I added a `pyupgrade --py37-plus` to the pre-commit. So, This PR title was changed a "Add PyUpgrade Pre-Commit Hook for support python 3.7 and above" 😄

.pre-commit-config.yaml
```diff
@@ -3,4 +3,6 @@ repos:
     rev: stable
     hooks:
     - id: black
-      args: [--check]
\ No newline at end of file
+      args: [--check]
+    - id: pyupgrade
+      args: [--py37-plus]
```
cclauss, line 8: Add one and only one carriage return to the end of the file so GitHub removes the ⛔ .

docs/autogen.py
```diff
@@ -1,9 +1,6 @@
-# -*- coding: utf-8 -*-
 # This file has been taken from Keras' `docs` module found here:
 # https://github.com/keras-team/keras/blob/master/docs/autogen.py
 #
-from __future__ import print_function
-from __future__ import unicode_literals
```
cclauss, line 6: Lines 10 thru 14 can also be removed.

.pre-commit-config.yaml
marload: complete :smile:

docs/autogen.py
marload: complete :smile:

cclauss (approved):
LGTM!

marctorsoc:
Got late to this one, thanks everyone!

PR #759, opened 2020-10-27
Travis CI: Add Python 3.9 to the testing

cclauss:
Removes the stages to allow all four jobs can run in parallel.

~Scipy is not yet compatible with Python 3.9 so run that job in `allow_failures` mode.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ jobs:
       python: 3.8
     - name: "Run tests on python 3.9"
       python: 3.9
-      env: NUMPY_VERSION=1.19.2
+      env: NUMPY_VERSION=1.19.3
   allow_failures:
     - name: "Run tests on python 3.9"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,6 @@ jobs:
     - name: "Run tests on python 3.9"
       python: 3.9
       env: NUMPY_VERSION=1.19.4
-  allow_failures: 
-    - name: "Run tests on python 3.9"
 
 install:
   - sudo apt-get remove ipython || true  # ipython is not pre-installed in Focal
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,33 +18,36 @@ env:
...
-    - stage: "test python 3.7"
-      name: "Run tests on python 3.7"
-      script: pytest
-      python: 3.7
-    - stage: "test python 3.8"
-      name: "Run tests on python 3.8"
-      script: pytest
-      python: 3.8
+before_script:
+  - sleep 15 # mongo takes time to start
```
marload, line 47: In my experience, it often takes more than 15 seconds to install mongodb. It seems better to increase this time to 30seconds.

cclauss: mongodb should be kicked off on line 19 and then we pip install a lot of stuff and that takes more than a full minute...

Look at the times along the right margin of the Travis output...
Screenshot 2020-10-27 at 14 51 20

marctorsoc:
> Removes the stages to allow all four jobs can run in parallel.
> 
> Scipy is not yet compatible with Python 3.9 so run that job in `allow_failures` mode.

I'm not sure about supporting python 3.9 if scipy is not compatible with it. Isn't better to wait until that happens rather than the allow_failures hack?

marctorsoc:
> > Removes the stages to allow all four jobs can run in parallel.
> > Scipy is not yet compatible with Python 3.9 so run that job in `allow_failures` mode.
> 
> I'm not sure about supporting python 3.9 if scipy is not compatible with it. Isn't better to wait until that happens rather than the allow_failures hack?

@cclauss ?

cclauss:
Scipy now has Python 3.9 wheels and these tests pass on Py39.

marctorsoc (approved):
LGTM, thanks!

PR #764, opened 2020-11-05
Cache dependencies pre-commit workflow

cclauss:
Cache the `pip` and `pre-commit` downloads. This will also cache all environments that `pre-commit` initializes at the start of each run.


marload:
LGTM!

marload (approved):

Issue #769, opened 2020-12-03
Choosing parameter space directly on objective function


brunodoamaral:
Hi all!

This is just a suggestion about a different way to use the minimization API. Here we annotate the search space directly on parameters.

I'm sure this may not be the best approach for all situations, but at least for me seems to be easier to link parameters and search space:

```python
# define an objective function
def objective(
            x: hp.uniform('x', -10, 10),
            y: hp.uniform('y', -10, 10)
        ):
    return (x * y) ** 2
```

We'll need to use the inspect module to get the annotations from function parameters:

```python
import inspect

# define a search space
space = inspect.getfullargspec(objective).annotations
```

And finally wrap the objective function on a dict -> kargs helper function:

```python
def objective_fmin_wrapper(func):
    def _objective(params):
        return func(**params)
    
    return _objective

# minimize the objective over the space
best = fmin(objective_fmin_wrapper(objective), space, algo=tpe.suggest, max_evals=1000)

print(best)
```

I'm sorry if this is a duplicate of future features. Anyway, I hope this can enrich the discussion about hyperopt.

Best regards,
Bruno

maxpumperla:
@brunodoamaral thank you, great suggestion. @marload I think you wanted to look into API going forward?

marload:
@maxpumperla 
To me this design looks very attractive. It cannot be replaced in all cases, but it seems that it can be used as an intuitive design in many cases. I will make related PR sooner or later.

@brunodoamaral 
Or do you have any plans to create your own PR?

brunodoamaral:
@marload and @maxpumperla thanks for your positive reception. I can create a PR, but first we should decide the design of this API.

One that come to my mind is:

```python

# define the objective function as before

# minimize the objective over the space
best = fmin(objective, space='auto', algo=tpe.suggest, max_evals=1000)
```
Within `fmin`, I will call `inspect.getfullargspec(objective).annotations` if `space == 'auto'` to get the search space. It should raise an Exception if the objective function has no annotations.

What do you think?

marload:
@brunodoamaral 

This design looks good to me. In addition, if the parameter obtained `inspecto.getfullargspec` is not Hyperopt Search Space, an error must be raised.

brunodoamaral:
@marload I just forked the master branch, but I'm having errors on tests. Do you know if they are ok?

Here is one of the errors:

```
======================================================================
ERROR: test_plot_vars (tests.test_plotting.TestPlotting)
[...]

PR #777, opened 2020-12-30
Initial version with support for parameter annotation on objective fn.

brunodoamaral:
Proposed implementation of https://github.com/hyperopt/hyperopt/issues/769.


marload (approved):
Great! LGTM

hyperopt/exceptions.py
```diff
@@ -38,4 +38,12 @@ class AllTrialsFailed(Exception):
     """All optimization steps have finished with status base.STATUS_FAIL"""
 
 
+class InvalidAnnotationParameter(ValueError):
```
marctorsoc, line 41:
```suggestion
class InvalidAnnotatedParameter(ValueError):
```
?

hyperopt/exceptions.py
```diff
@@ -38,4 +38,12 @@ class AllTrialsFailed(Exception):
...
+    """fn has a type hint that is not from hp."""
+
+    def __init__(self, result):
+        ValueError.__init__(self)
```
marctorsoc, line 45: cool way to init, didn't see this before 💯

hyperopt/fmin.py
```diff
@@ -97,6 +98,17 @@ def partial(fn, **kwargs):
     return rval
 
 
+def __objective_fmin_wrapper(func):
+    """
+    Wrap the objective function on a dict to kargs
```
marctorsoc, line 103: the typical way to abbreviate this is args and kwargs, but not kargs

hyperopt/fmin.py
```diff
@@ -506,6 +520,21 @@ def fmin(
...
 
+    if space == "auto":
+        # Read space from objective fn
+        space = inspect.getfullargspec(fn).annotations
+
+        # Validate space
+        for var, hp_func in space.items():
+            if not isinstance(hp_func, pyll.base.Apply):
+                raise exceptions.InvalidAnnotationParameter(
+                    f"Parameter {var} has type {hp_func} which is not "
```
marctorsoc, line 531: given it's a parameter, what if we call it `param` ?

marctorsoc: Either if we make "auto" the default value for `space` or not, I would give more info in this msg error, e.g:

```
When using `space="auto"`, please annotate the objective function arguments with a `pyll.base.Apply` subclass. See example in `fmin` docstring
```

hyperopt/tests/test_fmin.py
```diff
@@ -372,3 +372,21 @@ def test_early_stop_no_progress_loss():
...
+    def objective(x: hp.uniform("x", -10, 10), y: hp.uniform("y", -10, 10)):
+        return (x * y) ** 2
+
+    trials = Trials()
+    fmin(objective, space="auto", algo=tpe.suggest, max_evals=10, trials=trials)
+
+    assert len(trials) == 10
+
+
+@nose.tools.raises(exceptions.InvalidAnnotationParameter)
```
marctorsoc, line 387: can we do this in pytest? we're trying to move away (very slowly) from nose

marctorsoc (commented):
Really cool idea, great job! it's clean and very powerful change. Some general comments:

* I would document it a bit more. Not sure if we can add an example to the docs to make it clearer. See the great work by @marload here https://github.com/hyperopt/hyperopt/tree/master/docs/templates/tutorials
* Do we want space="auto" to be the default?
* What do you think about "annotated" rather than "auto"? the latter is a bit too generic IMO

hyperopt/fmin.py
brunodoamaral: Indeed... I'll update the code

hyperopt/fmin.py
brunodoamaral: Great suggestion

hyperopt/tests/test_fmin.py
brunodoamaral: I'll change this as requested.

brunodoamaral (commented):
Resolved in commit https://github.com/hyperopt/hyperopt/pull/777/commits/8a48244c33704c828980a77fe2896dfd5cf1e2c4

brunodoamaral:
@marctorsoc regarding your considerations:

 - I'll append to the tutorial notebook a few cells doing the same optimization, but with annotated space.
 - _Do we want space="auto" to be the default?_ - I would suggest to do it after a few months of usage. If the community thinks it is a good idea, we could leave it as default.
 - _What do you think about "annotated" rather than "auto"? the latter is a bit too generic IMO_ - indeed, "auto" sounds like we doesn't need to provide the space information.

At last, but not least, I'd like to provide an argument against my onw implementation: one may think that [type hint](https://docs.python.org/3/library/typing.html) is being misused in this approach. IMO it was designed to help the linter during development... not to incorporate information during runtime. Anyway, I still think it is a good idea to declare parameters and space in the same location.

marctorsoc:
> @marctorsoc regarding your considerations:
> 
> * I'll append to the tutorial notebook a few cells doing the same optimization, but with annotated space.

👍  let me know when it's finished and ready to review

> * _Do we want space="auto" to be the default?_ - I would suggest to do it after a few months of usage. If the community thinks it is a good idea, we could leave it as default.

fair enough

> * _What do you think about "annotated" rather than "auto"? the latter is a bit too generic IMO_ - indeed, "auto" sounds like we doesn't need to provide the space information.

great, let's go with "annotated" then. I had the same feeling, that if it's auto you don't have to do anything. And that's misleading
 
> At last, but not least, I'd like to provide an argument against my onw implementation: one may think that [type hint](https://docs.python.org/3/library/typing.html) is being misused in this approach. IMO it was designed to help the linter during development... not to incorporate information during runtime. Anyway, I still think it is a good idea to declare parameters and space in the same location.

the 3 maintainers of this repo approved either here or in the issue you opened, so you have the greenlight :)

marctorsoc:
> > @marctorsoc regarding your considerations:
> > 
> > * I'll append to the tutorial notebook a few cells doing the same optimization, but with annotated space.
> 
> 👍 let me know when it's finished and ready to review
> 
> > * _Do we want space="auto" to be the default?_ - I would suggest to do it after a few months of usage. If the community thinks it is a good idea, we could leave it as default.
> 
> fair enough
> 
> > * _What do you think about "annotated" rather than "auto"? the latter is a bit too generic IMO_ - indeed, "auto" sounds like we doesn't need to provide the space information.
> 
> great, let's go with "annotated" then. I had the same feeling, that if it's auto you don't have to do anything. And that's misleading
> 
> > At last, but not least, I'd like to provide an argument against my onw implementation: one may think that [type hint](https://docs.python.org/3/library/typing.html) is being misused in this approach. IMO it was designed to help the linter during development... not to incorporate information during runtime. Anyway, I still think it is a good idea to declare parameters and space in the same location.
> 
> the 3 maintainers of this repo approved either here or in the issue you opened, so you have the greenlight :)

is this ready to review now? @brunodoamaral

brunodoamaral:
Sorry for slow response. It went out of my radar these days.
Here is my proposal of first tutorial update: https://github.com/brunodoamaral/hyperopt/blob/master/tutorial/01.BasicTutorial.ipynb

maxpumperla:
@brunodoamaral I love what you did there, great improvement. I think your test still says "auto", which is probably why travis fails. Can you have another closer look and let me know when you're good to go with this?

marctorsoc:
thanks for your changes @brunodoamaral . It's a bit strange that we have /tutorials and then /docs/templates/tutorials. We should reorganise this at some point... 

anyways, there seems not possible (or at least doesn't work for me) to render these ipynb files in github and also revieweing ipynb files is not easy. Do you mind using jupytext? we can keep both the .py and .ipynb, or just the .py, I'm not strongly opinionated on this

brunodoamaral:
@marctorsoc I had no problems rendering the .ipynb on github (but sometimes I have to click on "reload" if failed). I don't think duplicating the code as a .py file will help... as the .ipynb has a lot more information, such as charts and rich text.

marctorsoc:
> @marctorsoc I had no problems rendering the .ipynb on github (but sometimes I have to click on "reload" if failed). I don't think duplicating the code as a .py file will help... as the .ipynb has a lot more information, such as charts and rich text.

I'm not opposed to keep .ipynb files. My point is that in PRs we shouldn't review .ipynb but .py. If you want to keep them both, I'm happy with it and it's really easy using jupytext. 

In other OSS projects like sklearn, they keep only the .py, and they have a build process to generate the .ipynb automatically, see https://github.com/scikit-learn/scikit-learn/blob/8ea176ae0ca535cdbfad7413322bbc3e54979e4d/.binder/postBuild#L27. What I propose makes your life easier, I think, as you can develop your notebook in jupyter, and push both files

marload:
I want to merge this PR. Is there any progress?
cc. @brunodoamaral

marctorsoc:
it's fine. Let's merge and do the jupytext stuff in some other iteration

brunodoamaral:
@marctorsoc thanks for the help. I was not familiar with jupytext and actually didn't understand how to procedure with your request. I will watch the changes made by the team to get a better picture on how to use it.

PR #782, opened 2021-02-08
Fixed implementation of "last_job_timeout" option

skylogic004:
Fixed implementation of "last_job_timeout" which is supposed to exit the mongo worker after some time has passed, but instead it was doing nothing.

Before, the condition in the if statement was impossible to evaluate to True. Now, the condition is evaluated every time a job is about to be reserved.


maxpumperla:
thank you for this PR @skylogic004

PR #811, opened 2021-09-07
Make hyperopt SparkTrials support pyspark 3.2 which pin thread mode enabled by default

WeichenXu123:
Make hyperopt SparkTrials support pyspark 3.2 which pin thread mode enabled by default

See https://github.com/apache/spark/pull/32429 for details.

### DIFF
--- a/hyperopt/tests/test_spark.py
+++ b/hyperopt/tests/test_spark.py
@@ -504,14 +504,6 @@ class FMinTestCase(unittest.TestCase, BaseSparkContext):
                     log_output=log_output
                 ),
             )
-            self.assertNotIn(
-                "Task should have been cancelled",
-                log_output,
-                """ "Task should have been cancelled" should not in log:
-                              {log_output}""".format(
-                    log_output=log_output
-                ),
-            )
             self.assert_task_failed(log_output, 0)
 
         # Test mix of successful and cancelled trials.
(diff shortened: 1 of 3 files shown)


HyukjinKwon (commented):
The changes look making sense to me from my side if tests pass

WeichenXu123:
Run test against spark 3.2 by command:
```
pytest hyperopt/tests/test_spark.py
```
It passed.

hyperopt/tests/test_spark.py
```diff
@@ -504,14 +504,6 @@ def fn(x):
...
                 ),
             )
-            self.assertNotIn(
-                "Task should have been cancelled",
-                log_output,
-                """ "Task should have been cancelled" should not in log:
-                              {log_output}""".format(
-                    log_output=log_output
-                ),
-            )
```
WeichenXu123, line 514: This line test is meaningless and it will fail under spark 3.2, so I remove it.

The log will print a line like:
```
trial task {tid} cancelled, exception is {e}
```

The exception we get from spark 3.2 is the original exception `Exception("Task should have been cancelled")`, but in spark < 3.2 it will get an wrapped exception which make the log does not contains message "Task should have been cancelled", this difference make the line test fail under spark 3.2.
But, this line test is meaningless, we only need to test the job was cancelled correctly, so I remove this line.

WeichenXu123:
@maxpumperla @marload Could you help review ? Thanks!

WeichenXu123:
Gentle ping @maxpumperla @marload

maxpumperla (approved):
LGTM

PR #814, opened 2021-09-27
Added plot script

NormanTUD:
This adds a nice script to plot parameters from a DB. It's not perfect in terms of code, but it works well enough to be usable to others I think.


maxpumperla:
@NormanTUD I'm not entirely convinced we should just add this without further consolidating what we already have.

There's a `plotting.py` and a `graph_viz.py` in the codebase for (not entirely) similar purposes. Adding a `plot.py` to the mix will likely be confusing. Maybe creating a `plot` module could be an idea?

NormanTUD:
> @NormanTUD I'm not entirely convinced we should just add this without further consolidating what we already have.
> 
> There's a `plotting.py` and a `graph_viz.py` in the codebase for (not entirely) similar purposes. Adding a `plot.py` to the mix will likely be confusing. Maybe creating a `plot` module could be an idea?

We may rename it, but imho, both of these don't really help in choosing limits for hyperparameters and don't give a real overview of the tested points and their correlation. In my kind of plot, you can easily see which parameters are highly correlated and which are not, for example, which is not possible in the original scripts.

maxpumperla:
I'm not at all arguing the usefulness of your addition, it's really cool! I just don't see how we have a `plotting.py` and a `plot.py` next to each other without being completely confusing to users.

I have no insights into how (and if) current users utilise what we have. It'd be difficult to justify deleting the old stuff.

NormanTUD:
Nobody needs to delete stuff. May we just rename it to "2d-scatterplot.py"? I really love HyperOpt and use it daily, but the tools provided with it, imho, don't really provide any useful insight. At least they never did for me. Therefore, I really think people would appreciate this. 
If you think it would not be a good idea for not confusing users, we may take the module route. But I'd really appreciate it if this tool would be provided to users somehow.

PR #805, opened 2021-05-17
Add hint regarding `rstate` optional param for `fmin`

LeowWB:
As issue #775 points out, the docs don't mention anything about random seed. Added a hint to fix that. Also updated a section reference that seems to be a leftover from the old wiki.

PR #821, opened 2021-10-30
Use numpy.random's Generator instead of legacy RandomState

rharish101:
Numpy's `numpy.random.RandomState` is deprecated. According to [its documentation](https://numpy.org/doc/stable/reference/random/legacy.html):
> This generator is considered frozen and will have no further improvements. It is guaranteed to produce the same values as the final point release of NumPy v1.16.

I've updated the code to use instances of `numpy.random.Generator` instead.

**NOTE**: This will lead to an API breakage, since `numpy.random.Generator` is not backwards-compatible with `numpy.random.RandomState`.

I've also updated some tests to use fixed RNG algorithms where the seed is important. However, I may have missed some, so please let me know in your review. The reason is that `numpy.random.Generator` doesn't provide a compatibility guarantee, as written in [its documentation](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator):
> **No Compatibility Guarantee**
>
> Generator does not provide a version compatibility guarantee. In particular, as better algorithms evolve the bit stream may change.

### DIFF
--- a/docs/templates/getting-started/minimizing_functions.md
+++ b/docs/templates/getting-started/minimizing_functions.md
@@ -84,7 +84,7 @@ date-times, you'll be fine.
 them as attachments.
 
 **HINT:** If you need to replicate the results of the stochastic search (e.g. for a demonstration),
-pass an object of type `np.random.RandomState` into the `fmin` function, using the `rstate` optional parameter.
+pass an object of type `np.random.Generator` into the `fmin` function, using the `rstate` optional parameter.
 
 Writing the function above in dictionary-returning style, it
 would look like this:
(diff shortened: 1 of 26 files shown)


rharish101:
BTW, why aren't the pytest tests set up in the CI? Why only pre-commit?

maxpumperla:
@rharish101 good question, we've had that for a very long time. I'd assume @marctorsoc did some changes to the pipeline. Can we activate tests again?

marctorsoc:
I didn't change anything... the last PR were it ran was #805 and the first one where it didn't is #810. Doesn't seem to be something related to our code, but it'd be great if @egregorimar can fix it 🙏

maxpumperla:
@marctorsoc yeah, I didn't want to accuse you of anything haha. I just remembered that you were the one pushing for black. Before that we didn't have this hook in CI, so something must have changed.

@marload any insights?

marctorsoc:
@maxpumperla in the PR I linked, black and tests are running. So I doubt black is the problem. It seems to be like the stage is not added to the CI, not sure why. Not familiar with Travis, but there's no code change from 805 to 810...

PR #824, opened 2021-11-04
DOC missing low parameter in hp.randint

egregorimar:
The documentation of `hp.randint` is missing the optional `low` parameter.


marctorsoc:
I didn't change anything... the last PR were it ran was https://github.com/hyperopt/hyperopt/pull/805 and the first one where it didn't is https://github.com/hyperopt/hyperopt/pull/810. Doesn't seem to be something related to our code, but it'd be great if @egregorimar can fix it 🙏

Issue #829, opened 2021-11-19
AttributeError with numpy while using HyperoptEstimator


HarshSharma12:
I tried using the `HyperoptEstimator` from hpsklearn and it gives me the following error 

```
estim.fit(X_train, y_train)
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hpsklearn/estimator.py", line 798, in fit
    fit_iter.send(increment)
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hpsklearn/estimator.py", line 708, in fit_iter
    return_argmin=False, # -- in case no success so far
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hyperopt/fmin.py", line 555, in fmin
    trials_save_file=trials_save_file,
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hyperopt/base.py", line 688, in fmin
    trials_save_file=trials_save_file,
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hyperopt/fmin.py", line 586, in fmin
    rval.exhaust()
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hyperopt/fmin.py", line 364, in exhaust
    self.run(self.max_evals - n_done, block_until_done=self.asynchronous)
  File "/home/harsh/anaconda3/envs/myenv/lib/python3.7/site-packages/hyperopt/fmin.py", line 279, in run
    new_ids, self.domain, trials, self.rstate.integers(2 ** 31 - 1)
AttributeError: 'numpy.random.mtrand.RandomState' object has no attribute 'integers'

```

According to the [numpy documentation on RandomState](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState), there is no attribute `integers`. There is however a `randint` which seems to satisfy the requirements.

Library versions:
- `hpsklearn @ git+https://github.com/hyperopt/hyperopt-sklearn@c66f8df76003b5e72cb242399eeb139f951a1883`
- `hyperopt==0.2.7`
- `numpy==1.21.4`

OS: Ubuntu 20.04

To replicate the error you can try the [Complete example using the Iris dataset:](https://github.com/hyperopt/hyperopt-sklearn/tree/master#usage) from the documentation

mandjevant:
@rharish101

rharish101:
[...]

PR #831, opened 2021-11-27
Bump numpy to earliest version that supports Generator

rharish101:
This is in reference to [this comment](https://github.com/hyperopt/hyperopt/issues/829#issuecomment-978914377) on #829:

> Hi. I just had the same problem.
> Wouldn't pinning or giving a range of the numpy version in `setup.py` prevent this? In a sense that contributors would not accidentally use an old numpy version.

_Originally posted by @alexandschaefer in https://github.com/hyperopt/hyperopt/issues/829#issuecomment-978914377_

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setuptools.setup(
     keywords="Bayesian optimization hyperparameter model selection",
     include_package_data=True,
     install_requires=[
-        "numpy",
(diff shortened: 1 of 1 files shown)


maxpumperla:
@rharish101 thanks, that's a good solution.

PR #832, opened 2021-11-29
Remove travis and unpin pyspark version in CI

WeichenXu123:
Signed-off-by: Weichen Xu <weichen.xu@databricks.com>

Remove travis and unpin pyspark version in CI


WeichenXu123:
CC @maxpumperla 
Thanks!

Issue #669, opened 2020-05-20
Hyperopt using Apache Spark  & broadcasted datasets


easygold:
Dear all,

We are trying to use the hyperopt library ver. 0.2.4 withApache Spark on a Big Data Cluster (Cloudera on premise)
In order to execute the hyperopt library with Apache Spark, we use as suggested the "SparkTrials" function,

example:
	 `trials = SparkTrials(parallelism=2, spark_session=spark)`

and the related "fmin" function, in order to get the best params:

```
best = fmin(fn=lambda params: objective(X_train, y_train, params),
                space=space_params,
                algo=partial(tpe.suggest, n_startup_jobs=n_start),
                max_evals=evals+n_start,
                trials=trials,
                rstate=np.random.RandomState(SEED),
                show_progressbar=False,
                verbose=False)

```

Using the above configuration, the library works fine; 
as next step, in order to improve the performances and avoid multiple calls to the Spark driver (avoid bottleneck problem), we have tried to use broadcasted datasets in the objective/"fn" function, as in the example below:

```
# broadcast dataset with Spark Broadcast
X_train_br = sparkContext.broadcast(X_train)

...

def objective(X_train_br, y_train_br, params):
        '''objective function to minimize with hyperopt'''
        # load broadcasted dataset (python list of lists)
        X_train = X_train_br.value

        ...
```

but the point is that we are not able to load the broadcasted dataset inside the objective/"fn" function. In detail, We are facing with the following issue:

```
File "/tmp/.../conda-env/lib/python3.6/site-packages/hyperopt/fmin.py", line 482, in fmin
    show_progressbar=show_progressbar,
  File "/tmp/.../conda-env/lib/python3.6/site-packages/hyperopt/spark.py", line 270, in fmin
    raise e
  File "/tmp/.../conda-env/lib/python3.6/site-packages/hyperopt/spark.py", line 266, in fmin
    show_progressbar=show_progressbar,
  File "/tmp/.../conda-env/lib/python3.6/site-packages/hyperopt/fmin.py", line 504, in fmin
    show_progressbar=show_progressbar,
  File "/tmp/.../conda-env/lib/python3.6/site-packages/hyperopt/fmin.py", line 152, in __init__
    pickler.loads(msg)
  File "/tmp/.../conda-env/lib/python3.6/site-packages/pyspark/broadcast.py", line 45, in _from_id
    raise Exception("Broadcast variable '%s' not loaded!" % bid)
[...]

PR #856, opened 2022-05-02
Avoid attachment in SparkTrials

candalfigomoro:
Fix https://github.com/hyperopt/hyperopt/issues/669.

### Problem

The following code in https://github.com/hyperopt/hyperopt/blob/master/hyperopt/fmin.py#L160
```
        if self.asynchronous:
            if "FMinIter_Domain" in trials.attachments:
                logger.warning("over-writing old domain trials attachment")
            msg = pickler.dumps(domain)
            # -- sanity check for unpickling
            pickler.loads(msg)
            trials.attachments["FMinIter_Domain"] = msg
```
tries to add an attachment to Spark trials, since SparkTrials sets `self.asynchronous` to True (just like MongoTrials). The problem is that it relies on pickling `domain` and this is an issue if an object cannot be serialized by pickle (e.g. if we are using a Spark-broadcasted object).

Moreover, this code in https://github.com/hyperopt/hyperopt/blob/master/hyperopt/spark.py#L188
```
    def trial_attachments(self, trial):
        raise NotImplementedError("SparkTrials does not support trial attachments.")
```
seems to suggest that SparkTrials does not even support trial attachments. So why are we trying to add an attachment if SparkTrials does not support trial attachments?

### Solution

We add an additional `and not hasattr(self.trials, "_spark")` condition, so the current behavior doesn't change for MongoTrials, but we skip the problematic code for SparkTrials (`_spark` is an attribute within SparkTrials which stores the Spark session). This "hasattr" check is similar to other implemented checks such as this one: https://github.com/hyperopt/hyperopt/blob/master/hyperopt/fmin.py#L204

### DIFF
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -157,7 +157,7 @@ class FMinIter:
         self.rstate = rstate
         self.verbose = verbose
 
-        if self.asynchronous:
+        if self.asynchronous and not hasattr(self.trials, "_spark"):
             if "FMinIter_Domain" in trials.attachments:
                 logger.warning("over-writing old domain trials attachment")
             msg = pickler.dumps(domain)


candalfigomoro:
@maxpumperla @WeichenXu123

lu-wang-dl (approved):

lu-wang-dl (approved):

WenjieAtBayer:
Thanks for the fix. Could you release a new version?

PR #861, opened 2022-07-03
fix for issue 565: list index out of range

dmytro-mz:
fix for issue #565 IndexError in ATPE
ATPE optimizer uses native random module instead of numpy random and there is a difference in `randint` function logic.
Native random [includes](https://docs.python.org/3/library/random.html#random.randint) upper boundary, while numpy [excludes](https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html) it.


IngoStatworx:
Can someone help with this?

aggiesean:
I have the same issue with ATPE. Hope someone can help.

lu-wang-dl (approved):

PR #874, opened 2023-02-06
Remove call to numpy.warnings.filterwarnings

akx:
That shouldn't happen as a side effect, not to mention it doesn't work with NumPy 1.24 anyway.


akx:
Lint failure is fixed in #875.

lu-wang-dl (approved):

PR #875, opened 2023-02-06
Modernize CI a bit

akx:
* use https://github.com/pre-commit/action
* split logically to different workflow files
* use pip caching


hyperopt/fmin.py
```diff
@@ -28,7 +28,7 @@
         'Failed to load cloudpickle, try installing cloudpickle via "pip install '
         'cloudpickle" for enhanced pickling support.'
     )
-    import six.moves.cPickle as pickler
```
lu-wang-dl, line 31: Is this change intended?

lu-wang-dl: I see. You are dropping Python 2 support in this PR too.

lu-wang-dl (approved):
LGTM. Thanks for reformatting this.

hyperopt/fmin.py
akx: Python 2 support has actually been dropped earlier (see #876).

PR #876, opened 2023-02-06
Make package formally Python 3.7+ only

akx:
Python 2 support had been effectively dropped in #831 and https://github.com/hyperopt/hyperopt/commit/caa3c2425de055d885b59acf5bd921e5dae33eb0, so update metadata to reflect that.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[bdist_wheel]
-universal = 1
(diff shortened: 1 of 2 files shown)


akx:
Lint failure is fixed in https://github.com/hyperopt/hyperopt/pull/875.

jensenerik:
This should allow you to remove the dependency on the `future` library right?

PR #877, opened 2023-02-06
Require py4j only for Spark

akx:


akx:
Lint failure is fixed in https://github.com/hyperopt/hyperopt/pull/875.

lu-wang-dl (approved):

Issue #562, opened 2019-11-01
Adding stale-bot to close and lock issues older than X days without activity


maxpumperla:
~250 issues isn't too bad for a project this size, but we have some _really_ old issues that in many cases don't even make sense anymore. would be great to have a little help there, as many other projects do.

brethvoice:
@maxpumperla how about 366 days as a start for X (in case of leap years)?  That means people need to at least work on/check in on their issues yearly.

PR #890, opened 2023-06-18
Add action for closing stale PRs

michaelmior:
Fix for #562. Currently set to mark as stale after 120 days and then close after an addition 30 days of no activity. @maxpumperla what do you think?


lu-wang-dl:
This will be helpful to cleanup the old issues. Thanks.

lu-wang-dl (approved):

PR #898, opened 2023-08-20
Remove `future` dependency

### LINKED ISSUE #897: Remove `future` library dependency
Following on from https://github.com/hyperopt/hyperopt/pull/876, since Python 2 support is dropped the `future` library should not be needed anymore. This would require removing `past` imports too as that packages is bundled with `future`

rparini:
Closes #897

Most of the instances of `old_div` clearly involved floats anyway so the `old_div` was not doing anything. There's a few instances (see https://github.com/hyperopt/hyperopt/commit/8e16cf2ffd7849f672a890163198e239ce4eca3b) where I'm guessing it's correct to replace `old_div` with `/` but I'm less confident so would be good if someone more familiar with the codebase could double check.

I've replaced `basestring` with `(str, bytes)` since that's the direct replacement but maybe some of these should be just `str`.

### DIFF
--- a/hyperopt/anneal.py
+++ b/hyperopt/anneal.py
@@ -9,7 +9,6 @@ The simple (but not overly simple) code of simulated annealing makes this file
 a good starting point for implementing new search algorithms.
 
 """
-from past.utils import old_div
 import logging
 import numpy as np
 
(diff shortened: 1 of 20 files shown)


michaelmior:
Thanks!

Issue #906, opened 2023-11-11
BUG: Problem with ATPE and hierarchical space


stevetracvc:
I'm pretty sure this is a bug, and to me it seems huge. As in, can we really use atpe.suggest in this sort of instance?

In this dummy example, there are two parameters in the search space: parent and child. If parent is True, then there is a child value. If parent is False, then there is NO child value. Using tpe.suggest, every time parent has a value of 0, child has a value of None. Using atpe.suggest, every time a parent has a value of 0, _child is still given a value!_

If you use `space_eval` then you don't see the erroneous values for child, but they are stored in the Trials object (and file, if storing locally). To me, this seems like it would incorrectly influence the atpe.suggest model (eg, when parent is False, the errant values of child are still used in the optimization), but I'll admit that I have a hard time following the internals of hyperopt so maybe this error only manifests as incorrectly storing values in the trials object.

Here's the simple code to reproduce

```
from hyperopt import fmin, atpe, tpe, Trials, hp
import numpy as np


def test_hyperopt(algo):
    trials = Trials()
    max_evals = 10

    # dummy objective function
    def fn(space):
        return np.random.rand()

    # note that if False is chosen for parent, then there is no child value
    # if True is chosen for parent, then there is a child value
    space = hp.choice('parent', [
        (False, None),
        (True, hp.choice('child', [1, 2, 3])),
    ])

    best = fmin(fn, space, algo=algo, trials=trials, max_evals=max_evals)

    for i in range(max_evals):
        current_idxs = trials.trials[i]['misc']['idxs']
        current_vals = trials.trials[i]['misc']['vals']
        print(f"Parent: {current_vals['parent']} -> "
              f"Child: {current_vals['child']}")
        try:
            # make sure each trial's idxs values match the evaluation number
            assert current_idxs['parent'] == [i]
        except AssertionError:
            print(f"Assertion error: ")
            print(f"expected value: [{i}]")
            print(f"actual value: {current_idxs['parent']}")


test_hyperopt(tpe.suggest)
test_hyperopt(atpe.suggest)

```

and output

```
100%|██████████| 10/10 [00:00<00:00, 1651.89trial/s, best loss: 0.11396600560034376 @ 2023-11-11 19:43:29.507000]
Parent: [1] -> Child: [0]
Parent: [1] -> Child: [0]
Parent: [0] -> Child: []
Parent: [1] -> Child: [2]
Parent: [0] -> Child: []
Parent: [0] -> Child: []
[...]

PR #907, opened 2023-11-11
fix: idxs in Trials object shares new_ids

stevetracvc:
This fixes the issue of all atpe Trials objects idxs values being [0]. See #906 for extra info


stevetracvc:
Do you have an explanation for why you don't want this package to work properly? Why close this?

PR #929, opened 2024-10-15
Create daho.py

nikhilgupta0:
Steps Breakdown:
Dataset Preparation: We load the Boston Housing dataset (or any dataset of your choice) and split it into training and test sets. XGBoost Model Training: The model is trained using xgboost.train, with the evaluation metric being mean_squared_error. Search Space Definition: The search space for hyperparameters is defined using tune.loguniform (for learning rate eta) and tune.randint, tune.uniform for other parameters. Asynchronous Search (ASHA Scheduler): The ASHA scheduler is used to schedule trials asynchronously. It ensures the most promising trials get more resources while terminating underperforming ones early. Distributed Execution: ray.init() starts the Ray cluster, which manages distributed execution across available CPUs.

### DIFF
--- a/daho.py
+++ b/daho.py
@@ -6,22 +6,27 @@ import pandas as pd
 from sklearn.model_selection import train_test_split
 from sklearn.metrics import mean_squared_error
 
+
 # Step 1: Load the dataset
 def load_data():
     # Example with Boston Housing dataset (You can replace it with any dataset)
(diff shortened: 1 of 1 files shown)


michaelmior:
It's unclear what the purpose of adding this directly to hyperopt is. If you think there's something missing from existing examples, suggestions on updating the documentation would be appreciated.

PR #950, opened 2026-04-15
fix: handle numpy.int64 scalar in switch node eval (fixes #949)

### LINKED ISSUE #949: TypeError: switch argument was for hp.choice nodes with numpy >= 1.20
## Description
Running hyperopt with hp.choice raises a TypeError in pyll/base.py when the sampled index is returned as a numpy.int64 scalar rather than a Python int:

```
TypeError: ('switch argument was', 0)
  File "hyperopt/pyll/base.py", line 863, in rec_eval
    raise TypeError("switch argument was", switch_i)
```
## Root Cause
In hyperopt/pyll/base.py around line 859, the switch node handler calls .item() on the sampled value only when it is an np.ndarray:

```
if isinstance(switch_i, np.ndarray):
    switch_i = switch_i.item()
if isinstance(switch_i, int):   # fails for np.int64 bare scalars
    ...
else:
    raise TypeError("switch argument was", switch_i)

```
Prior to numpy 1.20, numpy.int64 inherited from Python's int, so the isinstance(switch_i, int) check passed silently. As of numpy 1.20, that inheritance was removed ([numpy 1.20 release notes](https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations)), and bare numpy.int64 scalars now fall through to the TypeError.

## Reproduction
```
import numpy as np
print(np.__version__)          # e.g. 1.26.4
print(isinstance(np.int64(0), int))  # False — this is the core issue

from hyperopt import fmin, tpe, hp, STATUS_OK

space = {'layers': hp.choice('layers', [1, 2, 3])}
fmin(fn=lambda p: {'loss': p['layers'], 'status': STATUS_OK},
     space=space, algo=tpe.suggest, max_evals=5)
# Raises: TypeError: ('switch argument was', 0)
```

## Fix
Extend the existing ndarray guard to also handle bare np.integer scalars, which have the same .item() method:
```
# Before:
                if isinstance(switch_i, np.ndarray):
                    switch_i = switch_i.item()
                if isinstance(switch_i, int):
```

```
# After:
                if isinstance(switch_i, np.ndarray):
                    switch_i = switch_i.item()
                if isinstance(switch_i, np.integer):
                    switch_i = int(switch_i)
                if isinstance(switch_i, int):
```
`np.integer` is the abstract base class for all numpy integer scalar types (`np.int64`, `np.int32`, etc.) and has been stable across numpy versions. This is a one-line change with no behavioral impact for environments where `numpy.int64` still behaves as a Python int.

Environment
numpy >= 1.20
hyperopt @ c49ad148 (current master)
Python 3.11

Happy to open a PR with this fix if it looks good to maintainers.

jedisomCVNA:
## Summary

Fixes #949.

`numpy.int64` is not a subclass of Python `int` since numpy 1.20 ([numpy 1.20 release notes](https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations)), so the bare `isinstance(switch_i, int)` check in the `switch` node evaluator raises `TypeError` when the sampled index is a bare numpy integer scalar:

```
TypeError: ('switch argument was', 0)
  File "hyperopt/pyll/base.py", line 863, in rec_eval
    raise TypeError("switch argument was", switch_i)
```

The existing `np.ndarray` guard only handles zero-dimensional arrays via `.item()`. Bare `np.integer` scalars (which is what the TPE sampler returns) bypass it entirely.

## Fix

Add an explicit `np.integer` guard immediately after the `np.ndarray` check, coercing the scalar to a Python `int` before the existing `isinstance(switch_i, int)` check:

```python
# Before:
if isinstance(switch_i, np.ndarray):
    switch_i = switch_i.item()
if isinstance(switch_i, int):

# After:
if isinstance(switch_i, np.ndarray):
    switch_i = switch_i.item()
if isinstance(switch_i, np.integer):
    switch_i = int(switch_i)
if isinstance(switch_i, int):
```

`np.integer` is the stable abstract base class for all numpy integer scalar types (`np.int64`, `np.int32`, etc.) and has been available across all numpy versions. No behavioral change for environments where this bug is not triggered.

### DIFF
--- a/hyperopt/pyll/base.py
+++ b/hyperopt/pyll/base.py
@@ -856,6 +856,8 @@ def rec_eval(
                 switch_i = memo[switch_i_var]
                 if isinstance(switch_i, np.ndarray):
                     switch_i = switch_i.item()
+                if isinstance(switch_i, np.integer):
+                    switch_i = int(switch_i)
                 if isinstance(switch_i, int):
                     if switch_i < 0:
                         raise ValueError("switch pos must be positive int", switch_i)


jaberg:
Thanks for taking the time to do this PR

PR #952, opened 2026-05-15
fix for numpy version > 2

erdogant:
fix for numpy version > 2


jaberg:
Thanks for writing this up. It conflicted with another change but yours was better so I incorporated it in 9834314, now in main branch.

Issue #943, opened 2025-09-10
Please build a new version of hyperopt and share it at pypip


bees4ever:
There are changes made in the main branch, which are not published to the python pypip. Can you please create a new build for it? That would be nice. Thank you.

KubaGRe:
@michaelmior @lu-wang-dl Hello, is it possible to release new version with python 3.12?

KubaGRe:
@maxpumperla Hello, could you please make a new release?

kaweantonio:
@michaelmior who has permission to publish to PyPI? 

I also like the suggestion from @CarliJoy to use a Github Workflow for Publishing.

ysheikh2:
@bjkomer  @marctorsoc  @archsyscall @maxpumperla @jaberg  would you happen to know who has PyPi publishing permissions?

michaelmior:
@kaweantonio Sorry for the delay. Unfortunately I do not know who has publishing permissions.

jaberg:
I might still technically be able to publish a package, but I haven't done it in probably over 10 years. My guess is that @maxpumperla uploaded that one. Publishing to pypi with a workflow sounds good. All of @CarliJoy 's suggestions seem good. 

@CarliJoy if you don't mind putting that PR together, I'll merge it and figure out how to make it auto-publish from main.

CarliJoy:
It will take me some time. I saw there was already a pyproject.toml Migration. I will do it within a month.

Given the project is rarely updated I propose to change to a [uv](https://docs.astral.sh/uv/) based workflow. This not only speeds up the CI/CD but also properly locks the depenencies. Preventing failures due to updated depenencies.

Would you be okay with this?

jaberg:
Astral looks interesting, I'm intrigued. Go for it, I look forward to seeing the change.

Also, speaking of it being updated rarely. An important step toward refreshing it could be for folks to comment on what they use it for. If folks on this thread have a minute or two, could you chime in on this hyperopt discussion thread? https://github.com/hyperopt/hyperopt/discussions/954

jaberg:
See Discussion #957, closing this issue. Thanks all!

jaberg:
(Thanks especially @CarliJoy , I learned a lot of good modern python stuff from your PR. Hyperopt hasn't seen such a major refresh in years!)

jaberg:
Hopefully addressed via https://github.com/hyperopt/hyperopt/discussions/959, reply there if there are further issues?

[...]

PR #947, opened 2026-02-28
fix: migrate to pyproject.toml

mys721tx:
Migrates to pyproject.toml since pkg_resources is deprecated in setuptools 82.0.0. Fixes #946.

PR #955, opened 2026-06-21
Make release ready

CarliJoy:
Make the project easier (to maintain).

After [pypi is configured](https://docs.pypi.org/trusted-publishers/adding-a-publisher/)  with *workflow name* `publish-to-pypi.yml` and *environment* `pypi` correctly the project can now simply be released by creating a github release.

The project was also migrated to uv so the pipeline now uses an lock file. This way changes in upstream packages won't break the tests.

see #943

Disclaimer: As visible the commits I used claude code to fix the tests and run some rather core like works. I checked the changed for the compatibility so far I could but I am now expert in this packages. Please validate [this commit](https://github.com/hyperopt/hyperopt/pull/955/commits/ec004538584e9cec8aafc2051d036a8810b13d92) again.
I can vouch for all other changes so.


CarliJoy:
I added ruff as replacement for pyupgrade and black. It also activates isort rules now.

CarliJoy:
While waiting for the workflow to be approved I took the liberty to create a nox file to run the test in parallel.
So people developing can test locally first (and fast).

jaberg:
I read through the changes, and took them for a spin, and the only thing I want to suggest is to clarify in the README that the syntax of e.g. `pip install -e '.[MongoTrials,ATPE,SparkTrials,dev]'` is still meant to work, and still does work, and is meant to continue to work.  The use of uv is also supported. @CarliJoy does that all sound true?

@michaelmior were you the last one to package this up? I'll leave this un-merged for a couple of days for comment, but otherwise I think it's a good step toward livening up the project.

In my quick local experiment I see that there are some test failures in a Dockerfile that looks like this (ignoring uv deliberately), but at least lots of tests pass, which is more than I can say for what's in main (which should be renamed as such), which fails to run any tests at all if CarliJoy's branch below is replaced with the main one.
```
FROM python:3.11-slim
                 
WORKDIR /app     
                 
# Install virtualenv
RUN pip install --no-cache-dir virtualenv
                 
# Create and activate virtualenv
RUN virtualenv venv
                 
# Set the virtualenv as the default Python
ENV PATH="/app/venv/bin:$PATH"
                 
RUN apt-get update && apt-get upgrade
RUN apt-get install -y git
                 
RUN git clone https://github.com/CarliJoy/hyperopt.git
WORKDIR /app/hyperopt
RUN git checkout make-release-ready
                 
                 
# Install packages in the virtualenv (not uv)
#RUN pip install .
RUN pip install -e '.[MongoTrials,ATPE,SparkTrials,dev]'
                 
RUN pip install pytest                                                                                                                                                         
RUN pytest .
```

For clarity: I mean that this PR is an improvement, not that CarliJoy is also responsible for fixing all the tests. That's future work.

jaberg:
@maxpumperla I tagged you in here as a no-good-deed-goes-unpunished sort of thing. Possibly you put in some of the most recent changes. What do you think of this packaging PR that brings in uv?

jaberg:
@marctorsoc as I just noticed that you merged #947 two weeks ago. What do you think of uv?

CarliJoy:
> I read through the changes, and took them for a spin, and the only thing I want to suggest is to clarify in the README that the syntax of e.g. pip install -e '.[MongoTrials,ATPE,SparkTrials,dev]' is still meant to work, and still does work, and is meant to continue to work. The use of uv is also supported. @CarliJoy does that all sound true?

Well this is true but in this case you loose the guarantee of the pinning of the `uv.lock`. Meaning, when running `pip install -e '.[MongoTrials,ATPE,SparkTrials,dev]'` the latest version of the packages is used. Which can and will lead to difference between local execution and CI/CD execution. It is not recommended but it will work.

UV can export an requirements.txt file that can be consumed be pip. This way you *could* also install the pinning using pip. But I don't recommend it committing/doing it. You want once source of truth for the pinning not two.
Also uv handles installing python for the users, making it **very** easy to actually test hyperopt against all supported python version.

Here a Dockerfile that export the requirements.txt and and uses it in a second stage. I also had some test failing first but all of them, because using the `-slim` variant some libraries were missing.

```Dockerfile
ARG PYTHON_VERSION=3.11

# Stage 1: Export requirements.txt for the dev group using uv
FROM ghcr.io/astral-sh/uv:debian-slim AS uv-export

ARG PYTHON_VERSION

RUN apt-get update && apt-get install -y --no-install-recommends \
    git \
    ca-certificates \
    && update-ca-certificates \
    && rm -rf /var/lib/apt/lists/*
RUN git clone --branch make-release-ready --depth 1 https://github.com/CarliJoy/hyperopt.git  /app/hyperopt
WORKDIR /app/hyperopt
RUN uv export --python ${PYTHON_VERSION} --group dev --no-emit-project --frozen -o requirements.txt

# Stage 2: Install depenecies and hyperopt 
FROM python:${PYTHON_VERSION}-slim

WORKDIR /app

# git/ca-certificates: needed to clone the repo below over https
# libgomp1: required by the "lightgbm" python module (imported transitively
#   via hyperopt's atpe.py). lightgbm's compiled extension (lib_lightgbm.so)
#   is linked against GCC's OpenMP runtime, which python:slim doesn't ship.
#   Documented by LightGBM itself under "Custom Installation", "Linux" :
#   "In some rare cases, when you hit OSError: libgomp.so.1: cannot open
#   shared object file ... during importing LightGBM, you need to install
#   OpenMP runtime library separately (use your package manager and search
#   for lib[g|i]omp for doing this)."
#   See: https://pypi.org/project/lightgbm/
# default-jre-headless: pyspark's tests (test_spark.py) launch a local Spark
#   session, which needs a JVM. Without it, the Java gateway process can't
#   start. See: https://spark.apache.org/docs/latest/api/python/getting_started/install.html#dependencies
RUN apt-get update && apt-get install -y --no-install-recommends \
    git \
    ca-certificates \
    libgomp1 \
    default-jre-headless \
    && update-ca-certificates \
    && rm -rf /var/lib/apt/lists/*
RUN python -m venv /app/venv   # virtualenv is not required, venv is built-in in python 3.x
ENV PATH="/app/venv/bin:$PATH"

RUN git clone --branch make-release-ready --depth 1 https://github.com/CarliJoy/hyperopt.git  /app/hyperopt
WORKDIR /app/hyperopt

COPY --from=uv-export /app/hyperopt/requirements.txt .

# Install all dependencies
RUN pip install -r requirements.txt  
# Install project in editable mode, without depenencies
RUN pip install --no-deps -e '.'

CMD ["pytest", "-v", "hyperopt"]
```

Execute with
```bash
$ docker build -t hyperopt-testing - < Dockerfile
$ docker run hyperopt-testing
```


I also added an additional commit documentation the extras.  This should also trigger a CI/CD Pipeline. I think due to a bug in github one was lost after my last force push.

marctorsoc:
@jaberg we should really move to uv, it is much better. That being said, I'm not convinced of one PR doing everything vs just pip -> uv, and then another one for the 1M changes due to ruff, etc. If something goes wrong, it won't be easy to revert with everything packed here.

also, just skimmed but not sure how the version is calculated. This could be a major version change? Not sure

CarliJoy:
@marctorsoc I planned on making the ruff change within a second PR. But pre-commit didn't work and I wanted a working PR. They are still different commits.

EDIT: If wanted I could first create an update Python Version/Use ruff PR and then base a uv MR on it. But I am starting travelling tomorrow for a month. So this won't be done in a longer time.

jaberg:
@marctorsoc I can put together a changelog if @CarliJoy doesn't get to it before he travels, and you and I can review it and decide on a new version string? (and version string - mechanism as well if need be)

My TODO list on this PR to pick things up while @CarliJoy travels:
* [ ] draft changelog / release notes
* [ ] review how versioning works in hyperopt, choose a version
* [ ] enable the necessary workflow / actions in the GH settings (incl. share tokens with PyPI etc.)

jaberg:
(I should add, I'm busy this week and travelling myself next week, so it'll still take a bit to wrap this up.)

PR #960, opened 2026-09-08
fix: honor explicit progress controls independently of verbose

### LINKED ISSUE #937: Verbose=False and progress bar
Hi there team,

I hope you are all well, I am looking to have verbose=false in min to avoid output like 
TPE using 1/1 trials with best loss inf
but I want to keep the progress bar

Looking at the documentation it looks like if I set  verbose=false it will disable the progress bar too, is there a way to keep the progress bar but not have the verbose output?

DivyamTalwar:
Fixes #937

## The problem

`verbose` and `show_progressbar` are documented as separate controls, but `FMinIter` combined them:

```python
if not show_progressbar or not verbose:
    self.progress_callback = progress.no_progress_callback
```

Because `show_progressbar` defaulted to `True`, `verbose=False` won that `and` every time. Anyone who asked for a progress bar explicitly — or supplied a custom callback factory — got silence, with nothing in the output to explain why:

```python
# Before: no progress bar, despite asking for one.
fmin(objective, space, algo=tpe.suggest, max_evals=100,
     verbose=False, show_progressbar=True)

# Same for a custom reporter.
fmin(objective, space, algo=tpe.suggest, max_evals=100,
     verbose=False, show_progressbar=my_progress_factory)
```

The existing docstring already conceded the coupling ("If False, disable progress bar irrespectively of show_progressbar argument"), so the only way to combine quiet logging with a progress bar was to not use `verbose` at all.

## The fix

`show_progressbar` becomes tri-state, with `None` as the new default:

| value | behavior |
| --- | --- |
| `None` (default) | follows `verbose` — unchanged from today |
| `True` / `False` | authoritative, regardless of `verbose` |
| context-manager factory | authoritative, regardless of `verbose` |

`FMinIter` resolves `None` to `verbose` before the existing dispatch, so the only condition left to check is the caller's actual intent.

## Backward compatibility

The historical behavior is what `None` now means, so **no existing call changes**:

- `fmin(..., verbose=False)` → still no progress bar.
- `fmin(..., verbose=True)` → still a progress bar.
- Passing `show_progressbar` explicitly now does what its name says.

`None` is normalized in `fmin` *before* dispatching to a custom `Trials.fmin`, so the new sentinel never reaches duck-typed `Trials` implementations outside this repo that only understand a bool. That normalization is the reason the change touches `base.py` as well — `Trials.fmin`'s signature default had to move in step with the public one.

## Docs

`fmin`, `FMinIter`, and `Trials.fmin` now state the `None` default and spell out the `initial` / `total` keyword contract a custom factory has to satisfy, which was previously only discoverable by reading `hyperopt/progress.py`.

## Tests

`hyperopt/tests/unit/test_progress.py` gains coverage for the omitted default, explicit `True` under `verbose=False`, explicit `False` under `verbose=True`, a custom factory under `verbose=False`, and the custom-`Trials` dispatch path that must never see `None`.

## Verification

| check | result |
| --- | --- |
| `uv run pre-commit run --all-files` | pass |
| `uv run pytest hyperopt/tests/unit/test_fmin.py hyperopt/tests/unit/test_progress.py -q` | 24 passed |
| `PYSPARK_PIN_THREAD=true uv run pytest` | 343 passed, 36 skipped |
| `PYSPARK_PIN_THREAD=false uv run pytest hyperopt/tests/integration/test_spark.py -q` | 16 passed, 1 skipped |

Not covered locally: other Python versions and platform matrices — left to CI.

### DIFF
--- a/hyperopt/base.py
+++ b/hyperopt/base.py
@@ -638,7 +638,7 @@ class Trials:
         pass_expr_memo_ctrl=None,
         catch_eval_exceptions=False,
         return_argmin=True,
-        show_progressbar=True,
+        show_progressbar=None,
         early_stop_fn=None,
         trials_save_file="",
     ):
@@ -656,8 +656,10 @@ class Trials:
             error jobs (JOB_STATE_ERROR).  If set to False, such exceptions
             will not be caught, and so they will propagate to calling code.
 
-        show_progressbar : bool or context manager, default True.
-            Show a progressbar. See `hyperopt.progress` for customizing progress reporting.
+        show_progressbar : None, bool or context-manager factory, default None
+            Show a progress bar. When None, this follows the value of `verbose`.
+            A custom factory must accept `initial` and `total` keyword arguments.
+            See `hyperopt.progress` for customizing progress reporting.
 
         """
         # -- Stop-gap implementation!
--- a/hyperopt/fmin.py
+++ b/hyperopt/fmin.py
@@ -124,14 +124,16 @@ class FMinIter:
         timeout=None,
         loss_threshold=None,
         verbose=False,
-        show_progressbar=True,
+        show_progressbar=None,
         early_stop_fn=None,
         trials_save_file="",
     ):
         self.algo = algo
         self.domain = domain
         self.trials = trials
-        if not show_progressbar or not verbose:
+        if show_progressbar is None:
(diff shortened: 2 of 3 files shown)


jaberg:
Heya thanks for the PR! I can take a look later this week.

